PR #2, opened 2019-10-11
MANTA-4579 properly mark all skips and errors

rjloura:


src/jobs/mod.rs
```diff
@@ -165,6 +210,105 @@ impl Default for Job {
...
+// There are many reasons why an object may fail on a rebalance job.  This
+// enum of reasons is meant to be shared between both the agent and the
+// zone, and also apply to all future job actions.  We could build out
+// reasons for each job and then build translators for each agent
+// error/reason to each job action reason, but it would be nice to avoid
+// having to do that.  The trade off is that some of these reasons may be unique
+// to a specific job action. It's a trade off and we may decide in the future to
+// split these off but for now it seems to make the most sense going forward to
+// put them all in the same spot.
+//
```
rhb2: If I'm understanding this correctly (I may not be), something in the comment above made it sound to me like we're mixing object failure codes with job failure codes?  If that's true, I'm not sure that's a good idea.  I'm not saying that I disagree with it as of now -- I'd need to think about it more, but I am concerned about it leading to ambiguity of where a real failure occurred.

src/jobs/mod.rs
```diff
@@ -140,12 +152,32 @@ impl Task {
...
+        hasher.input(random_string(g, len).as_bytes());
+        let md5checksum = hasher.result();
+        let md5sum = base64::encode(&md5checksum);
+
+        Task {
+            object_id: Uuid::new_v4().to_string(),
+            owner: Uuid::new_v4().to_string(),
+            md5sum,
+            source: MantaObjectShark::arbitrary(g),
+            status: TaskStatus::arbitrary(g),
```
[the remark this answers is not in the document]
rjloura: In the `assignment_processing_test()` I actually override the arbitrary value added here. But it can't hurt the add the functionality, right?

src/jobs/mod.rs
rjloura: This is more about allowing the codes below to be used by the agent as well as with other Jobs that we implement in the future.  Presumably most (if not all) jobs will be operating on Objects, so I figured it would be good to put these in a common place for all future Jobs to use.

As far as the agent goes, the idea here is that the agent and the job_action can share this enum and thus eliminate the need for any agent_failure_reason => job_action_object_failure_reason translator.

rjloura: This is more about allowing the codes below to be used by the agent as well as with other Jobs that we implement in the future. Presumably most (if not all) jobs will be operating on Objects, so I figured it would be good to put these in a common place for all future Jobs to use.

As far as the agent goes, the idea here is that the agent and the job_action can share this enum and thus eliminate the need for any agent_failure_reason => job_action_object_failure_reason translator.

src/jobs/mod.rs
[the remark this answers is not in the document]
rhb2: Sounds good.

src/agent.rs
```diff
@@ -604,24 +604,21 @@ fn download(
...
+            return Err(ObjectSkippedReason::NetworkError);
         }
     };
 
     // If the response status code is anything other than 200 (Ok), flag
     // failure and return.
     if response.status() != reqwest::StatusCode::OK {
-        return Err(format!(
-            "Failed request with status: {}",
-            response.status()
```
rhb2: I am a little apprehensive about stripping away the actual status code of the response to a request that failed.  Presumably it could fail due to a number of different reasons, 404 only being one of them, but anyway, knowing that this could be anything, do we want to propagate that information back?

src/agent.rs
```diff
@@ -1083,14 +1080,17 @@ mod tests {
     //              as "Failed("Failed request with status: 404 Not Found")".
     #[test]
     fn object_not_found() {
```
rhb2: The name of the test should probably change as well as the comments above it describing what it means, since we're not specifically checking for a 404 anymore.  That being said, I still think it's important to distinguish between the different kinds of failures (i.e. != 200) that we can get in the response to a request that the agent sends to a mako.  We may not be able to anticipate them all of the failures now, but I believe we need to reserve the ability to be able to test for any of them.

src/agent.rs
rjloura: I added some resolution here.  LMK what you think.

rjloura:
> The other thing that I'd check on is what a database looks like (see /var/tmp/rebalancer/completed/) after processing an assignment where we know that some of the objects have failed. It looks like everything should work, but part of the underlying stats structure is now changing since the Task itself is changing and each stats structure has a list of failed tasks. I can help walk through verifying these parts too when you have time.

```

```sqlite> select * from tasks;
abc|rebalancer|QXBlX0QFcscVIwptkUaI8g==|dc|localhost|{"Failed":"SourceUnreachable"}
sqlite> select * from stats;
{"state":{"Complete":[{"object_id":"abc","owner":"rebalancer","md5sum":"QXBlX0QFcscVIwptkUaI8g==","source":{"datacenter":"dc","manta_storage_id":"localhost"},"status":{"Failed":"SourceUnreachable"}}]},"failed":1,"complete":1,"total":1}

sqlite> select * from stats;
{"state":{"Complete":[{"object_id":"area_codes_by_state.csv","owner":"rebalancer","md5sum":"abc","source":{"datacenter":"dc","manta_storage_id":"localhost"},"status":{"Failed":"MD5Mismatch"}}]},"failed":1,"complete":1,"total":1}
sqlite> select * from tasks;
area_codes_by_state.csv|rebalancer|abc|dc|localhost|{"Failed":"MD5Mismatch"}
sqlite>

File: src/agent.rs (hunk 4 above)
Comment (rjloura) [reply src/agent.rs]: done

File: src/agent.rs [hunk 5]
```diff
@@ -598,46 +598,70 @@ fn verify_file_md5(file_path: &str, csum: &str) -> bool {
     result_ascii == csum
 }
 
+fn _skip_download_error(
```
rhb2: I don't think this is necessary given that the agent should just log all errors regardless.  613 is my sad excuse of a logging statement which is addressed in my latest PR.  The other issue I have with this approach is that it essential boils all failures down to two possible types.  In my opinion, it's one thing to to distill it down to one of two different types because it's convenient when handling errors and we believe that we will handle all failures one of two ways, but we should be more explicit in what we log.  I think that what the actual response status code is matters.

src/agent.rs
```diff
@@ -1076,21 +1100,24 @@ mod tests {
...
     // Description: Attempt to download an object from a storage node where
-    //              the object does not reside.
+    //              the object does not reside will cause a client error.
     // Expected:    TaskStatus for all tasks in the assignment should appear
-    //              as "Failed("Failed request with status: 404 Not Found")".
+    //              as "Failed("SourceClientError")".
     #[test]
-    fn object_not_found() {
-        let reason = "Failed request with status: 404 Not Found".to_string();
+    fn client_error() {
```
rhb2: A test for client error is good, but I wanted to keep the test for a missing object.  I don't think those are the exact same thing...

src/jobs/evacuate.rs
```diff
@@ -2158,6 +2159,7 @@ mod tests {
 
     #[test]
     fn assignment_processing_test() {
+//        pretty_env_logger::init();
```
rhb2: Delete?

PR #4, opened 2019-10-17
MANTA-4527 Want common logging mechanism that includes thread name

rhb2:
MANTA-4610 Provide Bunyan-formatted logging for rebalancer

### DIFF
--- a/src/jobs/evacuate.rs
+++ b/src/jobs/evacuate.rs
@@ -17,6 +17,7 @@ use crate::jobs::{
 };
 use crate::moray_client;
 use crate::picker::{self as mod_picker, SharkSource, StorageNode};
+use crate::util;
 
 use std::collections::hash_map::Entry::{Occupied, Vacant};
 use std::collections::HashMap;
@@ -34,7 +35,7 @@ use crossbeam_deque::{Injector, Steal};
 use libmanta::moray::{MantaObject, MantaObjectShark};
 use moray::client::MorayClient;
 use reqwest;
-use slog::{o, Drain, Logger};
+use slog::{Level, Logger};
 use threadpool::ThreadPool;
 
 // --- Diesel Stuff, TODO This should be refactored --- //
@@ -304,11 +305,7 @@ impl EvacuateJob {
         db_url: &str,
     ) -> Self {
         let manta_storage_id = from_shark.into();
-        let plain = slog_term::PlainSyncDecorator::new(std::io::stdout());
-        let log = Logger::root(
-            Mutex::new(slog_term::FullFormat::new(plain).build()).fuse(),
-            o!("build-id" => "0.1.0"),
-        );
+        let log = util::create_bunyan_logger(std::io::stdout());
         let conn = SqliteConnection::establish(db_url)
             .unwrap_or_else(|_| panic!("Error connecting to {}", db_url));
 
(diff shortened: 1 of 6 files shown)


rhb2:
FYI, I am fine with holding off and merging in after you push your change.  Just wanted to get the review started -- I'm sure you'll have questions.

src/jobs/evacuate.rs
```diff
@@ -1331,17 +1334,18 @@ where
         match assign_rx.recv() {
             Ok(assignment) => {
                 {
-                    info!(
-                        "Posting Assignment: {}\n",
-                        serde_json::to_string(&assignment)?
+                    log!(Level::Info,
+                        "Posting Assignment: {:#?}\n",
```
rjloura: nice

src/jobs/evacuate.rs
```diff
@@ -946,20 +947,17 @@ fn start_sharkspotter(
...
 
-    debug!("Starting sharkspotter thread: {:?}", &config);
+    log!(Level::Debug, "Starting sharkspotter thread: {:?}", &config);
 
-    let plain = slog_term::PlainSyncDecorator::new(std::io::stdout());
-    let log = Logger::root(
-        Mutex::new(slog_term::FullFormat::new(plain).build()).fuse(),
-        o!("build-id" => "0.1.0"),
-    );
+    let log = util::create_bunyan_logger(std::io::stdout());
```
rjloura: is it possible to use the same logger that was created in main.rs instead of creating a new one here?

src/jobs/evacuate.rs
```diff
@@ -304,11 +305,7 @@ impl EvacuateJob {
         db_url: &str,
     ) -> Self {
         let manta_storage_id = from_shark.into();
-        let plain = slog_term::PlainSyncDecorator::new(std::io::stdout());
-        let log = Logger::root(
-            Mutex::new(slog_term::FullFormat::new(plain).build()).fuse(),
-            o!("build-id" => "0.1.0"),
-        );
+        let log = util::create_bunyan_logger(std::io::stdout());
```
rjloura: Do we still need this now that we have the global logger?

src/jobs/evacuate.rs
```diff
@@ -449,7 +446,8 @@ impl EvacuateJob {
             .join()
             .expect("Sharkspotter Thread")
             .unwrap_or_else(|e| {
-                error!("Error joining sharkspotter handle: {}\n", e);
+                log!(Level::Error, "Error joining sharkspotter handle: {}\n",
```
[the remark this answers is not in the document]
rhb2: Old habits die hard... :)

src/jobs/evacuate.rs
rhb2: This doesn't even get used at all.  It's an artifact of an earlier implementation.  It should be removed.  Thanks for catching that.

src/jobs/evacuate.rs
rhb2: Strictly speaking, we don't, however you had mentioned that you wanted to be able to pass a logger to other crates at some point from the evacuate.rs code which is why I kept this one in there.  The authors of slog_scope strongly advise against using the "slog_scope" logger in libraries.  Granted, that's not the same thing as another crate, but it could be.

We could still remove this and legitimately clone the global logger and pass that on to another crate if need be.  If you agree with that, then I'll yank the line in 308.

rjloura: I think that makes sense.  Not sure.  Ideally I'd like to pass this Logger (or a child of it) to other crates that require a Logger (i.e. sharkspotter and rust-moray)

src/jobs/evacuate.rs
rjloura: I see.  We clone the global Logger on 958 when we call `sharkspotter::run()`, right?

PR #8, opened 2019-10-25
MANTA-4629 add Jenkinsfiles to Manta repositories

trentm:


trentm:
Note: failure of the one check was because I canceled the redundant branch build.

PR #11, opened 2019-10-30
MANTA-4655 Rebalancer agent should protect against receiving duplicate assignmen…

rhb2:
…ts in rapid succession

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -7,7 +7,7 @@
 /*
  * Copyright 2019, Joyent, Inc.
  */
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::ffi::OsString;
 use std::fs;
 use std::fs::File;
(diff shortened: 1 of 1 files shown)


src/agent.rs
```diff
@@ -78,19 +78,55 @@ struct PathExtractor {
 #[derive(Clone, Debug)]
 pub struct Agent {
     assignments: Arc<Mutex<Assignments>>,
```
rjloura: Does this hashmap only contain the assignments that are to be processed?

src/agent.rs
```diff
@@ -78,19 +78,55 @@ struct PathExtractor {
...
+    quiescing: Arc<Mutex<HashSet<String>>>,
```
rjloura: The intent for this hashmap is to hold the assignment in memory until it can be written out to disk, thus ensuring that an assignment with the same uuid isn't received while the write to disk is happening.  Is that correct?

src/agent.rs
```diff
@@ -78,19 +78,55 @@ struct PathExtractor {
...
 
     pub fn run(addr: &'static str) {
         info!("Listening for requests at {}", addr);
         gotham::start(addr, router());
     }
+
+    // Given an assignment uuid, check for its presence in both the "scheduled"
+    // and "completed" directory.  If found in either, or the assignment is in
+    // the process of being saved to disk, return true, otherwise false.
+    fn assignment_exists(&self, uuid: &str) -> bool {
```
rjloura: might be worth noting upfront (or renaming this function) so that it is clear if the assignment is not found it will be added to the quiecing set.  In the future if someone reads this code (or code where this is called) they may not realize that the quiecing set is could be altered.

src/agent.rs
rhb2: Exactly.  Moreover, I did not want to fence the entire operation (i.e. saving to disk) with a mutex because of the amount of time it would take.  Thus, I arrived at using a mutex for a quick memory check instead.

src/agent.rs
rhb2: Assignments are loaded in to *that* HashMap by workers at processing time.  This ensures that the total number of assignments in memory will not increase beyond what the agent is actively working on.  When a worker is free to process an assignment, it (supplied with the uuid of the assignment) does the loading in to the HashMap itself.  When the worker is done, it cleans up after itself.  That is, it removes the assignment from the HashMap and flushes it out to disk, currently in /var/tmp/rebalancer/completed/<uuid>.

To prefetch on a potential direction that this conversation is going in, I briefly considered overloading the functionality of that HashMap to handle the issue that I'm intending to address with this change, but I believe that the current approach is simplest logic for handling the problem.

src/agent.rs
rhb2: I thought I was pretty clairvoyant about that exact point on 122-124.  Given how small this function is, I don't see how someone could miss it.  Are you suggesting I move that comment to the top of the function?  Ultimately, this function should really only ever be called from one place (where it is now) which is part of why I didn't even declare it to be public.

rjloura: Even though it is not `pub` it could still be called from anywhere in this module.  Regardless, I just think a bit more clarity here would go a long way with little downside.  Ultimately though this is a nit, and everything else looks good here, so I'll +1 either way.

rjloura:
LGTM.

Lots of process changes going on with the transition from gerrit to GH PRs, but I think we are still putting the bug ID (issue ID) in the commit message, so just remember to update that when you merge.

PR #12, opened 2019-10-30
MANTA-4646 want alternate method for initial assignment generation

rjloura:


rjloura:
All unit tests pass including the new test implemented to specifically test this approach. I still need to test this on a deployed Manta, but am running afoul of MANTA-4655.  However, this is a fairly large change so I want to get it out there for review asap.

src/jobs/evacuate.rs
```diff
@@ -249,6 +251,11 @@ pub struct EvacuateJob {
     /// Maximum number of tasks to include in a single assignment.
     pub max_tasks_per_assignment: Option<u32>,
 
+    /// Maximum size of objects to include in a single assignment.
```
rhb2: I didn't see where it's being used so I don't have context.  Is this like maximum number of bytes per assignment?  If so, I think we need a better name for it.

rjloura: good catch.  Currently we limit the maximum size of an assignment based on the shark's available space.  I was thinking we should also limit the total assignment size by some configurable value, that is what this field was for.  I'll remove it and replace with a TODO statement for now since I don't want to increase the size of this wad any more.

rjloura:
After restarting the agents and redirecting their output to a file I was able to rebalance objects in my deployed datacenter.  Using the default limit the rebalancer ingested 9 objects from sharkspotter.  9 of 9 objects were successfully rebalanced as verified by the log.  I manually verified 3 of the 9 (the first two and the last one) via:
* `findobjects` on moray zone to verify `sharks` metadata
* `ls` on destination storage zone to verify objects were downloaded
* `mls` from my remote client to verify end to end reachability

src/jobs/evacuate.rs
```diff
@@ -1302,231 +1445,316 @@ where
...
-                            .len()
-                {
-                    shark_index = 0;
+            while !done {
+                // TODO: MANTA-4519
+                // get a fresh shark list
+                let mut shark_list =
+                    job_action.get_shark_list(Arc::clone(&picker), &algo)?;
+
+                // If sharks are busy remove them.
```
rhb2: Even though we already discussed this today, I think it's worth considering allowing busy sharks (agents?) to remain eligible.  Sending them more work that they can't immediately get to won't cause any real kind of memory bloat, so it should be safe.

src/jobs/evacuate.rs
```diff
@@ -1302,231 +1445,316 @@ where
...
+                    if shark_list.iter().any(|s| &s.manta_storage_id == key) {
+                        continue;
+                    }
+                    trace!("Telling thread to stop");
+                    remove_keys.push(key.clone());
                 }
+                _stop_join_some_assignment_threads(
+                    &mut shark_hash,
+                    remove_keys,
+                );
```
rhb2: If I'm understanding correctly, would this be necessary if we didn't "disqualify" busy agents?

src/jobs/evacuate.rs
```diff
@@ -1302,231 +1445,316 @@ where
...
+                            // That is to say, if any of the object's sharks
+                            // match the shark_hash_ent, then we can't use this
+                            // one for this object.
+                            // So !any is what we are looking for here.
+                            !eobj
+                                .object
+                                .sharks
+                                .iter()
+                                .any(|s| &s.manta_storage_id == *shark_id)
+                                */
```
rhb2: Since this is commented out, do we still need it?  The logic looks similar to validate_destination().

src/jobs/evacuate.rs
rjloura: We'd still need this code to cover the case where the `available_mb` on a shark decreases enough for it to fall out of the top X sharks (where X is some configurable number, currently set to 5 (see line 1476)).  A shark could also go offline during a rebalance, and this block of code would cover that case as well.

src/jobs/evacuate.rs
rjloura: Ah yes.  Good catch.  That is what I originally had when prototyping this, and replaced it with a call to `validate_destination`, which as you know is more comprehensive.

rjloura:
Let me rework the shark_busy stuff and see what happens when I throw multiple assignments at a shark at once.

rjloura:
Removed the `shark_busy` checks, but retained the shark state in case we need it for debugging.  I ran some unit tests as well as some initial tests on my dev environment and those look good.  QA should be testing this now.

rhb2:
> If there are no sharks available at the beginning of the job then I don't see a reason to spin waiting on sharks to become available.
> 
> It is possible that sharks could become unavailable or available during a job, so in that case I retry.
> 
> I was thinking about adding a retry threshold in here as well, but I wanted to keep this change smaller, and get it out for review. I can put a TODO statement in there for now.

I'm really trying to get clarification on one thing: If the job has not been started yet, it's ok to just completely give up, but if we're mid job and run in to the same situation, it's not ok to give up?

> Even if this is refactored, we'd still need to check if the list is empty and if so sleep and continue to the next iteration.
> 
> ```
> let mut shark_list = job_action.get_shark_list_with_log();
with_log()?

> if shark_list.is_empty() {
>   thread::sleep();
>   continue;
> }
> ```

Another thought I had is expanding the get_shark_list() interface (whether it takes an object that has a couple of members, or just two additional arguments) which contains a max number of retries and a delay time in between retries.

PR #14, opened 2019-11-04
MANTA-4678 Rebalancer agent unit tests cannot be run in parallel

rhb2:


rhb2:
> Looks good. Why use lazy_static?

Good question.  I had to learn something new about Rust to arrive at that point.  Originally, I tried just creating the variable as a static, i.e.:

static mut INITIALIZED: Mutex<bool> = Mutex::new(false);

But I kept getting a cryptic (at least to me) compiler error:

error[E0015]: calls in statics are limited to constant functions, tuple structs and tuple variants

What I arrived at through a lot of digging was that code required to initialize the structure doesn't run at compile time.  The use of lazy_static! takes care of that.  It runs the code to initialize the mutable static variable upon first access during runtime.  This alleviated my woes.  If there is a simpler way to do it, let me know, but to the best of my knowledge, this is the easiest/safest known way for something like this.

rhb2:
> Looks good. Why use lazy_static?

Good question.  I had to learn something new about Rust to arrive at that point.  Originally, I tried just creating the variable as a static, i.e.:

static mut INITIALIZED: Mutex<bool> = Mutex::new(false);

But I kept getting a cryptic (at least to me) compiler error:

error[E0015]: calls in statics are limited to constant functions, tuple structs and tuple variants

What I arrived at through a lot of digging was that code required to initialize the structure doesn't run at compile time.  The use of lazy_static! takes care of that.  It runs the code to initialize the mutable static variable upon first access during runtime.  This alleviated my woes.  If there is a simpler way to do it, let me know, but to the best of my knowledge, this is the easiest/safest known way for something like this.

PR #15, opened 2019-11-06
MANTA-4686 Rebalancer agent should create its own assignments for uni…

rhb2:
…t testing


src/agent.rs
```diff
@@ -1081,17 +1048,47 @@ mod tests {
...
+            tasks.push(task);
+        }
+
+        tasks
+    }
+
+    fn object_to_task(path: &Path) -> Task {
+        Task {
+            object_id: path.file_name().unwrap().to_str().unwrap().to_owned(),
+            owner: "rebalancer".to_owned(),
```
rjloura: This should be a UUID as a string.

src/util.rs
```diff
@@ -8,13 +8,32 @@
...
+use std::fs;
 use std::io;
 use std::sync::Mutex;
 use std::thread;
 
 use clap::{crate_name, crate_version};
+use md5::{Digest, Md5};
 use slog::{o, Drain, Logger};
 
+pub fn calculate_file_md5(file_path: &str) -> String {
```
rjloura: might be a good addition to https://github.com/joyent/rust-utils

src/agent.rs
rhb2: I deliberately chose to call it "rebalancer" instead of using a uuid in this case because I want the objects downloaded to go in /manta/rebalancer.  Visually, it's easier to tell it apart from other directories that may also exist in /manta.

Even though the test results should be sufficient when verifying whether or not a test succeeded and I trust them, I always verify by manually inspecting the contents of that directory at the completion of the test.  That's not a requirement, but if one wants to do it, I believe it should be easy.  The agent does not care what the owner name is anyway.

src/util.rs
rhb2: Yeah, I agree.  I could move it over and send a PR for that while this is on hold I guess.

PR #18, opened 2019-11-08
MANTA-4692 add logging to metadata update path

rjloura:


rjloura:
Oops.  This brought in changes from the job_status branch.

PR #22, opened 2019-11-13
MANTA-4724 Rebalancer agent should allow test code to supply its own custom function for processing tasks

rhb2:


src/agent.rs
```diff
@@ -918,83 +931,15 @@ mod tests {
...
-        let assignment: Assignment = match serde_json::from_str(&data) {
-            Ok(a) => a,
-            Err(e) => panic!(format!("Failed to deserialize: {}", e)),
-        };
-
-        assignment
-    }
-
     // Poll the server indefinitely on the status of a given assignment until
     // it is complete.  Currently, it's not clear if there should be an
```
rhb2: These two functions were moved to util.rs.  Other than that, no change to them except for that fact that they accept a &TestServer now as an argument.  In agent unit testing, they previously used the one that was globally declared, but now that the rebalancer zone can create a TestServer representing an agent, it may (and probably will) want to use more than one kind of agent (i.e. one that always succeeds, one that always fails, etc).

src/agent.rs
```diff
@@ -800,8 +811,9 @@ fn process_assignment(assignments: Arc<Mutex<Assignments>>, uuid: String) {
     assignment_complete(assignments, uuid);
 }
 
-/// Create a `Router`
-fn router() -> Router {
+// Create a `Router`.  This function is public because it will have external
+// consumers, namely the rebalancer zone test framework.
+pub fn router(f: fn(&mut Task)) -> Router {
```
rjloura: I like this approach of passing a callback to handle task processing.  I am curious though why expose the `router()` method instead of adding the same `fn(&mut Task)` arg to the already public `Agent::new()`?

I think the external difference would essentially be running the Agent in a separate thread instead of running a Gotham TestServer in a separate thread.  But I might be confused about that...

One fairly minor benefit of going through the `Agent::new()` method is that you have less possible external code paths to be concerned with when making an internal change.

rhb2: Not sure I follow.  For one thing, today, router() calls Agent::new(), not the other way around.  That aside, TestServer::new() requires an argument of type `Router`, therefore, no matter how you slice it, we have to pass an argument of type `Router` to TestServer::new().

Based on your commend about the external difference, it sounds like you're suggesting that we bail on using TestServer and instead just create an agent in a separate thread.  We could do that, but it would actually be more code for us and frankly it undermines the point of having TestServer anyway.  When running outside of test, as you can see, we pass the desired port to the agent when we start it.  In the case of the TestServer, it asynchronously fires up an agent and does so on a negotiated port (i.e. bind to port 0) returning what's basically a client to the caller.  That's what TestServer really is under the hood: a client which is aware of what port the newly started agent is listening on.  In this way, it allows for multiple agents to run within the same process without the possibility of a collision and it relieves the developer of having to maintain that information.  This way, you can have an "always pass agent" running at the same time as an "always fail" agent or an "honest agent".

To prefetch on something else, I'm not in favor of implementing an agent where you can dynamically switch out the process_task() function at run time without any assurance of knowing exactly where it is at a given point in time in its processing of an assignment.  I'm also not in favor of including information with an assignment that dictates which process_task() function to use.  I've been outspoken about this in the past, but in production, I don't even want something like that to be possible.  Then of course, you could suggest that we create different routes that eventually map to one process_task() or another, but I don't want to expose those end points in production code either.

Last but not least, we could ditch TestServer and create an agent like they currently do: it's a simple call to bind to 0 + a spawn and the retention of some information about the web server that we just started, but if we did all that, we've basically reimplemented TestServer.

src/jobs/evacuate.rs
```diff
@@ -2384,8 +2384,8 @@ mod tests {
         let mut rng = rand::thread_rng();
         let available_mb: u64 = rng.gen();
         let percent_used: u8 = rng.gen_range(0, 101);
-        let filesystem: String = util::random_string(rng.gen_range(1, 20));
-        let datacenter: String = util::random_string(rng.gen_range(1, 20));
+        let filesystem = util::test::random_string(rng.gen_range(1, 20));
```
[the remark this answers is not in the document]
rhb2: Yeah I had the same thought myself -- I wasn't sure why we have two different versions here in the first place.  No idea how it got that way, but I can clean that up for you.  My main justification for leaving it the way it is is that I'm not completely clear on what test requirements it would introduce so I didn't want to mess with something that I didn't completely understand despite the fact that the two functions appear to be identical in purpose.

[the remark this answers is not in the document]
rjloura: Eitherway is fine.  I think once it was brought in it was only used for new implementations.

src/agent.rs
rjloura: You're right, I misunderstood the hierarchy of agent.rs(mod) > router() > Agent.  What you say makes sense.

src/agent.rs
rhb2: I replaced all calls to util::random_string() with the rust-quickcheck-helper version with the aim of also retaining the functionality of having variable length.  Based on full_test, it appears to continue generating random strings of random lengths where it originally did.

PR #26, opened 2019-11-27
MANTA-4753 Rebalancer agent requires documentation

rhb2:


docs/agent.md
```diff
@@ -0,0 +1,313 @@
...
+The [Manta](https://github.com/joyent/manta) object rebalancer agent is a
+subsystem within the
+[Rebalancer project](https://github.com/joyent/manta-rebalancer) which is
+orchestrated by the rebalancer manager.
+
+# Overview
+The rebalancer agent is a service which runs on every storage node in a Manta
+deployment and its actions are orchestrated by a per-region manager referred to
+as the rebalancer "manager".  Its primary role is to handle requests containing
+an assignment posted by the manager.  The following interfaces are currently
```
rjloura: This last sentence seems out of place.  Perhaps this could move down to another heading just above where the interfaces are described.

docs/agent.md
```diff
@@ -0,0 +1,313 @@
...
+```
+
+```
+Job Management
+
+FLAGS:
+    -h, --help       Prints help information
+    -V, --version    Prints version information
+
+```
```
rjloura: Is this block intended for the manager?

docs/agent.md
```diff
@@ -0,0 +1,313 @@
...
+cargo check
+cargo clippy
+cargo test agent -- --nocapture | bunyan
+```
+
+## Testing
+
+### Execution
+As discussed in the development section above, anytime that a change is made to
+the agent code (or any subsystem that it diretly consumes), at a minimum, a
```
rjloura: "directly"

docs/agent.md
```diff
@@ -0,0 +1,313 @@
...
+to create additional instances of an agent, accessible only to that thread, if
+the agent that is globally accessible does not meet the needs of the test.  The
+primary reason for having additional agents would be if the developer wanted an
+agent that processed tasks in a different way than the way it would in
+production.  For example, an agent that blindly always fails tasks might be
+necessary if we are testing how the agent (or a client of the agent handles
+failure scenarios).
+
+### Development
+When developing test automation for the agent (or for the rebalancer zone, where
```
rjloura: Do we want to start referring to the "zone" as the "manager" going forward?

docs/agent.md
rhb2: It is.  This is one of the changes I need to make that I was alluding to in ~rebalancer.

docs/agent.md
rhb2: I need to remove the Job Management to something but I think it's wise to have those flags for the agent.  One other thing worth mentioning which you may have already gathered by now: I wrote this with the assumption that the agent is its own binary despite the fact that it's not yet.

docs/agent.md
rhb2: It'd be my preference, yeah.  It was actually in the process of drafting this document that the very subject had occurred to me.  There may be mixed usage of the terms throughout, here since we decided on manager like halfway through writing this.  This may not be the only place where I use zone.  I need to check for others.

docs/agent.md
rjloura: Yup.  I think that's fine.

PR #29, opened 2019-12-05
MANTA-4820 remora dies on storage nodes looking for postgres libraries when run as an agent

rhb2:


rhb2:
No real change to the code here, except to the Makefile and Cargo.toml so that the agent can be built without being dependent on pq-sys.so.x

PR #30, opened 2019-12-09
MANTA-4832 rebalancer should use pickers pagination

rjloura:


src/jobs/evacuate.rs
```diff
@@ -2306,21 +2306,21 @@ fn metadata_update_worker(
 
             info!("Assignment Complete: {}", &assignment.id);
 
-            if job_action
+            let removal = job_action
```
rjloura: This was changed because clippy was complaining.  Not sure how it sneaked in before.

PR #32, opened 2019-12-12
MANTA-4869 Rebalancer metadata update worker apparently lost conneciton, resulting in continuous errors during shrimp evacuation

rjloura:


rjloura:
This last commit includes a refactor as well as adds a newline to the job create response (so that the terminal prompt gets its own line), and cleans up some formatting elsewhere.

PR #39, opened 2019-12-20
MANTA-4916 rebalancer-adm: Short circuit job create functionality

rhb2:


src/main.rs
```diff
@@ -22,7 +22,14 @@ fn main() -> Result<(), Error> {
 
     match command.subcommand {
         SubCommand::Server => Ok(()),
-        SubCommand::DoJob(job) => job.run(),
+        SubCommand::DoJob(_job) => {
+            println!(
+                "This function is currently not allowed from the CLI. \
+                 Please contact the rebalancer-manager directly to create a \
```
rjloura: Can you say "Please contact the rebalancer-manager via its REST API on port 8888 to create a..."

rhb2: Done.

PR #45, opened 2020-01-14
MANTA-4950 rebalancer should use moray with updated cueball

rjloura:


rjloura:
I added MANTA-4886 to this as well because it is the reason we are updating cueball.  The change for MANTA-4886 essentially reflects a reversal of the workaround applied as part of MANTA-4891.

I've tested the cueball integration but need to verify that lwps are no longer orphaned/leaked with this fix.

rjloura:
Going to put this PR on hold.  It seems like this might not quite fix things.
MANTA-4966

src/util.rs
```diff
@@ -37,8 +37,11 @@ pub fn init_global_logger() -> slog_scope::GlobalLoggerGuard {
 pub fn get_thread_name() -> String {
     let cn = crate_name!().to_owned();
 
+    // If no name print the crate with the thread ID like so:
+    //      "<crate_name>[ThreadId(<tid>)]: <log msg>"
     if thread::current().name().is_none() {
-        return cn;
+        let ret = format!("{}[{:?}]", cn, thread::current().id());
+        return ret;
```
rhb2: Why not just include the thread id regardless and if the particular thread doesn't have a name, we can explicitly call it "unnamed" or something?

rjloura: Updated.

rjloura:
"MANTA-4886 lack of available lwps causes rebalancer to crash" was also fixed as part of this push

PR #52, opened 2020-02-12
MANTA-5022 Manta Rebalancer: Introduce metrics

rhb2:


rebalancer/src/metrics.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+impl Default for MetricLabels {
+    fn default() -> Self {
+        Self {
+            host: "127.0.0.1".into(),
+            port: 3020,
+            datacenter: "development".into(),
+            service: "1.rebalancer.localhost".into(),
+            server: "127.0.0.1".into(),
```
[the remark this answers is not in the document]
rhb2: I mentioned in our conversation earlier (in MM) that I thought it was desirable to implement the functionality for reading in the needed information from a config file as a separate ticket.    And yes, I think it could just as easily be a HashMap.  I don't necessarily see one approach as better than the other per say, but this is a set of information that will always be required and I think of a HashMap as something that is used when the set of information required could vary in size.

rebalancer/src/metrics.rs
```diff
@@ -0,0 +1,160 @@
...
+            request_count,
+            object_count,
+            error_count,
+        }
+    }
+}
+
+// Given the service configuration information, create (i.e. register) the
+// desired metrics with prometheus.
+pub fn register_metrics(labels: &MetricLabels) -> RegisteredMetrics {
```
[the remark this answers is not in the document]
rhb2: Actually, now that I think about it, it's not obvious to me how using a HashMap instead of this structure is that much of a win.  I think that same argument could be used for almost any structure that we have defined in the rebalancer.  This at least gives a reader an idea of what is required when the server starts.  I guess the argument could be made that I could also enumerate those requirements in comments, but I do feel like the approach here gives an explicit guarantee about what information is required and it's type.

Now that I think about it, with a HashMap, don't all values have to be of the same time?  So to have mixed types in there, wouldn't I have to define some kind of enum to get around this?

rebalancer/src/metrics.rs
```diff
@@ -0,0 +1,160 @@
...
+
+#[derive(Clone)]
+pub struct RegisteredMetrics {
+    pub request_count: CounterVec,
+    pub object_count: Counter,
+    pub error_count: CounterVec,
+}
+
+impl RegisteredMetrics {
+    fn new(
```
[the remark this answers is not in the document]
rhb2: Sorry, may I have to chock this up to probably not knowing about some rust syntactic sugar that you do.  Are you sayin we don't need the "new" function?  If so, I guess I would agree and I can eliminate it honestly.  I want to make sure I understand the question correctly before I do that though.

rebalancer/src/metrics.rs
```diff
@@ -0,0 +1,160 @@
...
+    // different kinds of possible errors that a given application could
+    // encounter and in the event that there are too many possibilities, only
+    // track certain error types and maintain the rest in a generic bucket.
+    let error_counter = register_counter_vec!(opts!(
+        "error_count",
+        "Errors encountered."
+    ).const_labels(const_labels.clone()), &["error"])
+    .expect("failed to register error_count counter");
+
+    RegisteredMetrics::new(
```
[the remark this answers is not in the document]
rhb2: Indeed.  I thought about that.  Then the "register_metrics" function could walk the list of these things and, depending on the value for "type" that you have there, we would call a certain prometheus register macro (there are lots of different ones depending on the type of metric), then take whatever is returned from it and wrap it in something else (I guess probably an enum) and then store it in a vector or HashMap that could be accessed at the time that we want to bump a count on one of them.  If we did something like that, then we could eliminate the "RegisteredMetrics" structure entirely and use a HashMap for that.  Ultimately, I chose the approach I have here as a first recourse because it's not obvious to me that the amount of work that it would take to make a system *that* robust would ever really pay for itself -- at least not in the context of the rebalancer.  If you think it's worth it, I'll do it.  I wanted to anyway.  I just thought it would be received as overkill.

rebalancer/Cargo.toml
```diff
@@ -23,6 +24,7 @@ joyent-rust-utils = { git = "https://github.com/joyent/rust-utils", tag = "v0.2.
 libmanta = { git = "https://github.com/joyent/rust-libmanta", tag = "v0.7.0" }
 mime = "0.3.13"
 md-5 = "0.8.0"
+prometheus = "0.5.0"
```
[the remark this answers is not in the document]
rhb2: Can use 0.7.0.

rebalancer/src/libagent.rs
```diff
@@ -799,13 +836,28 @@ fn process_assignment(
...
 // consumers, namely the rebalancer zone test framework.
+#[allow(clippy::many_single_char_names)]
 pub fn router(f: fn(&mut Task)) -> Router {
     build_simple_router(|route| {
+        let config = metrics::MetricLabels::default();
+        let metrics = metrics::register_metrics(&config);
+        let metrics_host = config.host.clone();
+        let metrics_port = config.port;
+
+        thread::spawn(move || {
```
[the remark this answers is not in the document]
rhb2: Done.

rebalancer/src/libagent.rs
```diff
@@ -413,11 +418,15 @@ fn post(agent: Agent, mut state: State) -> Box<HandlerFuture> {
                 // an assignent out of the message body.
```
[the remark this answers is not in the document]
rhb2: Yeah

rebalancer/src/libagent.rs
```diff
@@ -413,11 +418,15 @@ fn post(agent: Agent, mut state: State) -> Box<HandlerFuture> {
...
                     Ok(uv) => uv,
-                    Err(_e) => {
+                    Err(e) => {
                         let res = create_empty_response(
                             &state,
                             StatusCode::BAD_REQUEST,
                         );
+                        agent.metrics.lock().unwrap()
+                            .error_count
+                            .with_label_values(&[&e])
```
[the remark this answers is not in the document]
rhb2: Yes.  You'd want to do it from one of the prometheus servers.

rebalancer/src/metrics.rs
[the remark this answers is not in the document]
rjloura: Yes, I don't think we need a `new()` method.

rebalancer/src/metrics.rs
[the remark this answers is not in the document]
rjloura: I think a likely next step here would be to implement metrics for the manager, in which case we could use what you propose directly instead of having to modify anything in the metrics module.

rebalancer/src/metrics.rs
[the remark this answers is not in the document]
rjloura: I agree on the HashMap part. I think I got a bit turned around between RegisterMetrics and MetricLabels.  

I do however think it is worth while to add these fields to the config.json (via config-agent template).  Then register_metrics wouldn't need to take MetricLabels as an argument at all.  Or it could take an `Option<MetricLabels>` if you wanted to override them for testing or something.

rebalancer/src/metrics.rs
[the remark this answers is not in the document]
rjloura: See below.

rhb2:
No need to resume review yet.  Still have a couple things that I need to add.

agent/src/main.rs
```diff
@@ -52,7 +52,7 @@ fn main() {
     // that we process which will dictate the network on which to listen.  It is
     // worth mentioning that this will likely also be the case for the agent
     // port.
-    Agent::run("0.0.0.0:7878");
+    Agent::run(Some("config.json"));
```
rhb2: This needs to change to a config.toml, stored in /opt/smartdc/rebalancer-agent/etc.  Maybe I can add functionality to take a path (for testing purposes).  The work on the mako has already been done to render the config file and place it there though at least.

rhb2:
I see there are conflicts with Cargo.lock that still have to be resolved.  I'll handle that after we agree on the rest of this.  I changed the counter_*_inc() function to take a reference to a HashMap instead of cloning.  I also use a counter_vec for the object count now too and I've changed the API of counter_vec_inv() now such that the third argument (i.e. bucket) is an Option<&str>.  This way object count can indeed be categorized by something like an action in the manager while the agent can leave it as a single stat.

rebalancer/src/libagent.rs
```diff
@@ -74,27 +103,37 @@ struct PathExtractor {
...
             quiescing,
             tx,
+            metrics,
         }
     }
 
-    pub fn run(addr: &'static str) {
+    pub fn run(cfg_path: Option<&str>) {
+        let config = match cfg_path {
+            Some(c) => read_file(c),
```
rjloura: can we change this to `parse_config_file` or `agent_config_from_file`?

rebalancer/src/libagent.rs
```diff
@@ -410,14 +449,20 @@ fn post(agent: Agent, mut state: State) -> Box<HandlerFuture> {
         .then(move |full_body| match full_body {
             Ok(valid_body) => {
                 // Ceremony for parsing the information needed to create an
-                // an assignent out of the message body.
+                // an assigment out of the message body.
```
rjloura: "assignment"

rebalancer/src/metrics.rs
```diff
@@ -0,0 +1,187 @@
...
+// This enum exists so that we can take various prometheus counter types as the
+// same data type.  This is necessary so that we can store all metrics that we
+// create in the same hash map regardless of the type of counter.  Note, not
+// all prometheus counters are enumerated below.  Add them as needed.
+#[derive(Clone)]
+pub enum Metrics {
+    MetricsCounterVec(CounterVec),
+}
+
+#[allow(irrefutable_let_patterns)]
```
rjloura: where is the irrefutable `let` pattern in this?  Is it because `enum Metrics` only has `MetricsCounterVec`?

rebalancer/src/metrics.rs
```diff
@@ -0,0 +1,187 @@
...
+pub fn counter_vec_inc<S: ::std::hash::BuildHasher>(
+    metrics: &HashMap<String, Metrics, S>,
```
rjloura: can we change this to `metrics_map` so that it is not confused with `metrics` module, and `Metrics` enum?

rebalancer/src/metrics.rs
```diff
@@ -0,0 +1,187 @@
...
+    key: &str,
+    bucket: Option<&str>,
+) {
+    match metrics.get(key) {
+        Some(metric) => {
+            if let Metrics::MetricsCounterVec(c) = metric {
```
rjloura: I wonder if instead of creating a function for incrementing each type of metric (counter vs counter_vect, etc) we could just implement an `inc()` method on the `Metrics` enum like so:

```rs
impl Metrics {
  fn inc(&self) -> {
    match *self {
      Metrics::MetricsCounterVec(c) => {
        c.with_label_values(...).inc()
      },
      Metrics::<others> => {
         // do other inc methods
      }
    }
}
```

then the `counter_vec_inc()` function would then just be `metric_inc()` and internally call this other function that would do the right thing depending on the metric type.

rhb2: Given that counter_inc() and counter_vec_inc() take a different number of arguments, it's not obvious to me how we can have one API to rule them all for incrementing either type of counter.  Can you elaborate a bit more?  It could just be my misunderstanding of rust.

rebalancer/src/metrics.rs
rhb2: I think I'm going to leave this the way it is.  This variable name is very commonly used all over the place in manta when passing around a collector (which is essentially what this is).

rebalancer/src/metrics.rs
rhb2:
```
  --> rebalancer/src/metrics.rs:64:13
   |
64 | /             if let Metrics::MetricsCounterVec(c) = metric {
65 | |                 // Increment the total.
66 | |                 c.with_label_values(&["total"]).inc();
67 | |
...  |
73 | |                 }
74 | |             }
   | |_____________^
```

rebalancer/src/libagent.rs
rhb2: I think the way to go (since it's only returning a structure that's meaningful to the agent) is to move it in to the `impl` section for the agent itself.  Also, I don't think it needs to be public either.  I think I'll do that.

rebalancer/src/metrics.rs
```diff
@@ -112,21 +116,21 @@ pub fn register_metrics(labels: &ConfigMetrics) -> HashMap<String, Metrics> {
     .expect("failed to register incoming_request_count counter");
 
     metrics.insert(
-        "request_count".to_string(),
+        REQUEST_COUNT.to_string(),
```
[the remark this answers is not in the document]
rhb2: After thinking about it a bit, coupled with something else that you had asked about, I think we can hit two birds with one stone.  I've created a type alias

```
pub type MetricsMap = HashMap<&'static str, Metrics>;
```

which can be used all over the place as the type instead of the HashMap itself.  I think this will reduce the likelihood of a developer confusing a variable name in this case with it's data type.  Not to mention, if, for whatever reason, we change the key type again in the future, there will not be as many maintenance points.

rebalancer/src/metrics.rs
rjloura: Seems like the single field enum is why.  No explanation? 

No big deal.  I just haven't seen that warning before.

rebalancer/src/metrics.rs
rjloura: Was thinking there would be different ways to increment a given metric, but now I see that there are just different operations on a given metric.  I think what you have makes sense.

rhb2: I still don't know if I'm 100% tracking on your initial suggestion, but if you think it's possible to somehow do, it could be worth it.

rebalancer/src/metrics.rs
```diff
@@ -101,12 +103,12 @@ pub fn register_metrics(labels: &ConfigMetrics) -> HashMap<String, Metrics> {
     let request_counter = register_counter_vec!(
         opts!(REQUEST_COUNT, "Total number of requests handled.")
             .const_labels(const_labels.clone()),
-        &["op"]
+        &["req"]
```
rjloura: should we change the comment above as well?
> (e.g. op=GET, op=POST)

rebalancer/src/metrics.rs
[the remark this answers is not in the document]
rjloura: very nice!

rebalancer/src/metrics.rs
rhb2: Thanks, good catch.

PR #61, opened 2020-03-10
MANTA-5086 rename rebalancer-manager to rebalancer (smf and log upload names)

trentm:
This renames the SMF service name and log upload name to "rebalancer".
However, the *binary* name is still called "rebalancer-manager".


trentm:
Testing notes are on the ticket (https://jira.joyent.us/browse/MANTA-5086).

PR #62, opened 2020-03-11
MANTA-4942 rebalancer add ability to parse config options and tunables from config.json

rjloura:
MANTA-5037 clean up rebalancer config file location and generation

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,9 @@ members = [
 [profile.dev]
 panic = "abort"
 
+[profile.release]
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -8,6 +8,9 @@ members = [
 [profile.dev]
 panic = "abort"
 
+[profile.release]
+panic = "abort"
+
```
rjloura: This change is for "MANTA-5097 rebalancer should abort on thread panics in release bits"

PR #72, opened 2020-04-09
MANTA-5160 rebalancer should allow for static number of threads and a bounded queue

rjloura:
MANTA-5161 rebalancer status query could be more memory efficient
MANTA-5162 rebalancer only needs one gotham thread
MANTA-5164 rebalancer should remove skipped assignment from hash
MANTA-5168 rebalancer should use jemallocator 
MANTA-5169 rebalancer should use persistent reqwest clients 
MANTA-5100 rebalancer logging clean up

### DIFF
--- a/agent/Cargo.toml
+++ b/agent/Cargo.toml
@@ -20,7 +20,7 @@ libmanta = { git = "https://github.com/joyent/rust-libmanta", tag = "v0.7.0" }
 mime = "0.3.13"
 moray = { git = "https://github.com/joyent/rust-moray", tag = "v0.9.2" }
 rebalancer = { path = "../rebalancer" }
-reqwest = "0.9.18"
+reqwest = "0.9.24"
 rusqlite = "0.19.0"
 serde = { version = "1.0.91", features = ["derive"] }
 serde_derive = "1.0.91"
--- a/manager/src/config.rs
+++ b/manager/src/config.rs
@@ -44,6 +44,7 @@ pub struct ConfigOptions {
     pub max_assignment_size: usize,
     pub max_metadata_update_threads: usize,
     pub max_sharks: usize,
+    pub use_static_md_update_threads: bool,
 }
 
 impl Default for ConfigOptions {
@@ -52,6 +53,7 @@ impl Default for ConfigOptions {
             max_assignment_size: DEFAULT_MAX_ASSIGNMENT_SIZE,
             max_metadata_update_threads: DEFAULT_MAX_METADATA_UPDATE_THREADS,
             max_sharks: DEFAULT_MAX_SHARKS,
+            use_static_md_update_threads: false,
         }
     }
 }
--- a/manager/src/config.rs
+++ b/manager/src/config.rs
@@ -30,6 +30,7 @@ static DEFAULT_CONFIG_PATH: &str = "/opt/smartdc/rebalancer/config.json";
 static DEFAULT_MAX_ASSIGNMENT_SIZE: usize = 50;
 static DEFAULT_MAX_METADATA_UPDATE_THREADS: usize = 2;
 static DEFAULT_MAX_SHARKS: usize = 5;
+static DEFAULT_STATIC_QUEUE_DEPTH: usize = 10;
 
 #[derive(Deserialize, Default, Debug, Clone)]
 pub struct Shard {
@@ -45,6 +46,7 @@ pub struct ConfigOptions {
     pub max_metadata_update_threads: usize,
     pub max_sharks: usize,
     pub use_static_md_update_threads: bool,
+    pub static_queue_depth: usize,
 }
 
 impl Default for ConfigOptions {
@@ -54,6 +56,7 @@ impl Default for ConfigOptions {
             max_metadata_update_threads: DEFAULT_MAX_METADATA_UPDATE_THREADS,
             max_sharks: DEFAULT_MAX_SHARKS,
             use_static_md_update_threads: false,
+            static_queue_depth: DEFAULT_STATIC_QUEUE_DEPTH,
         }
     }
 }
--- a/manager/src/config.rs
+++ b/manager/src/config.rs
@@ -27,9 +27,23 @@ use std::thread;
 use std::thread::JoinHandle;
 
 static DEFAULT_CONFIG_PATH: &str = "/opt/smartdc/rebalancer/config.json";
+
+// The maximum number of tasks we will send in a single assignment to the agent.
 static DEFAULT_MAX_ASSIGNMENT_SIZE: usize = 50;
+
+// The maximum number of threads that will be used for metadata updates.
+// Each thread has its own hash of moray clients.
 static DEFAULT_MAX_METADATA_UPDATE_THREADS: usize = 2;
+
+// The maximum number of sharks we will use as destinations for things like
+// evacuate job.  This is the top 5 of an ordered list which could mean a
+// different set of sharks each time we get a snapshot from the storinfo zone.
 static DEFAULT_MAX_SHARKS: usize = 5;
+
+// The number of elements the bounded metadata update queue will be set to.
+// For evacuate jobs this represents the number of assignments that can be in
+// the post processing state waiting for a metadata update thread to become
+// available.
 static DEFAULT_STATIC_QUEUE_DEPTH: usize = 10;
 
 #[derive(Deserialize, Default, Debug, Clone)]
--- a/manager/src/config.rs
+++ b/manager/src/config.rs
@@ -135,6 +135,7 @@ impl Config {
                             .expect("Lock snaplink_cleanup_required");
 
                         *slcr = new_config;
+                        debug!("Configuration has been updated: {:#?}", *slcr);
                     }
                     Err(e) => {
                         warn!(
--- a/manager/src/jobs/evacuate.rs
+++ b/manager/src/jobs/evacuate.rs
@@ -646,6 +646,8 @@ impl EvacuateJob {
                 // TODO: should we panic? if so just replace with expect()
                 panic!("{}", e);
             });
+
+        self.remove_assignment_from_cache(assign_id);
     }
 
     // Removes assignment cache entry from cache.  The cache entry is
--- a/manager/src/jobs/evacuate.rs
+++ b/manager/src/jobs/evacuate.rs
@@ -22,7 +22,8 @@ use rebalancer::util::{MAX_HTTP_STATUS_CODE, MIN_HTTP_STATUS_CODE};
 
 use crate::config::{Config, ConfigOptions};
 use crate::jobs::{
-    Assignment, AssignmentCacheEntry, AssignmentId, AssignmentState, StorageId,
+    assignment_cache_usage, Assignment, AssignmentCacheEntry, AssignmentId,
+    AssignmentState, StorageId,
 };
 use crate::moray_client;
 use crate::pg_db;
@@ -597,8 +598,6 @@ impl EvacuateJob {
                 set_run_error(&mut ret, e);
             });
 
-        drop(job_action);
-
         ret
     }
 
@@ -676,9 +675,12 @@ impl EvacuateJob {
             format!("Remove assignment not in hash: {}", assignment_id)
         );
 
-        drop(entry);
-
+        let initial_size = assignment_cache_usage(&assignments);
         assignments.shrink_to_fit();
+        trace!(
+            "Assignment cache reduced by {}",
+            initial_size - assignment_cache_usage(&assignments)
+        );
     }
 
     fn skip_object(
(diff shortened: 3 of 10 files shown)


rjloura:
Just opening this up to start the conversation.  I might still have some changes.

manager/src/moray_client.rs
```diff
@@ -33,20 +33,16 @@ use resolve::{record::Srv, DnsConfig, DnsResolver};
 fn get_srv_record(svc: &str, proto: &str, host: &str) -> Result<Srv, Error> {
     let query = format!("{}.{}.{}", svc, proto, host);
     let r = DnsResolver::new(DnsConfig::load_default()?)?;
-    match r
-        .resolve_record::<Srv>(&query)?
+    r.resolve_record::<Srv>(&query)?
```
rjloura, line 36: No change in functionality here, just a more idiomatic way of doing things.

manager/src/main.rs
```diff
@@ -17,6 +17,11 @@ extern crate serde_derive;
 #[macro_use]
 extern crate rebalancer;
 
+// jemallocactor drastically improves our memory footprint
```
rjloura, line 20: MANTA-5168 rebalancer should use jemallocator

manager/src/jobs/status.rs
```diff
@@ -19,23 +19,35 @@ use std::string::ToString;
 
 use diesel::prelude::*;
 use diesel::result::ConnectionError;
+use diesel::sql_query;
```
rjloura, line 22: All the changes in this file are related to:
MANTA-5161 rebalancer status query could be more memory efficient

manager/src/jobs/mod.rs
```diff
@@ -347,28 +347,50 @@ impl Job {
 
         self.update_state(JobState::Running)?;
         debug!("Starting job {:#?}", &self);
-        println!("Starting Job: {}", &self.id);
+        info!("Starting Job: {}", &job_id);
```
rjloura, line 350: The only changes in this file are adding some logging messages to help us roughly track performance in the future.

manager/src/config.rs
```diff
@@ -30,6 +30,7 @@ static DEFAULT_CONFIG_PATH: &str = "/opt/smartdc/rebalancer/config.json";
 static DEFAULT_MAX_ASSIGNMENT_SIZE: usize = 50;
 static DEFAULT_MAX_METADATA_UPDATE_THREADS: usize = 2;
 static DEFAULT_MAX_SHARKS: usize = 5;
+static DEFAULT_STATIC_QUEUE_DEPTH: usize = 10;
```
rjloura, line 33: The changes in this file allow us to optionally use a static queue.  We may eventually decide to make this the default.
MANTA-5160 rebalancer should allow for static number of threads and a bounded queue

manager/src/jobs/evacuate.rs
```diff
@@ -414,6 +415,10 @@ pub struct EvacuateJob {
     /// Tunable options
     pub options: ConfigOptions,
 
+    pub post_client: reqwest::Client,
+
+    pub get_client: reqwest::Client,
+
```
rjloura, line 421: These are added for MANTA-5169 rebalancer should use persistent reqwest clients

manager/src/jobs/evacuate.rs
```diff
@@ -2243,6 +2258,28 @@ where
...
+                false
+            }
+            TryRecvError::Empty => {
+                trace!("No shutdown message, keep Checker running");
+                true
+            }
+        },
+    }
+}
+
```
rjloura, line 2282: This is a refactor.

manager/src/jobs/evacuate.rs
```diff
@@ -2455,10 +2473,78 @@ fn reconnect_and_reupdate(
     Some((robj, rclient))
 }
 
+enum UpdateWorkerMsg {
+    Data(AssignmentCacheEntry),
+    Stop,
+}
+
+fn metadata_update_worker_static(
```
rjloura, line 2481: This is where the bulk of the changes are for MANTA-5160 rebalancer should allow for static number of threads and a bounded queue

There's some additional refactoring below as well.

manager/src/main.rs
rhb2: |sed 's/mallocactor/mallocator/cg'

rhb2:
> To review "MANTA-5160 rebalancer should allow for static number of threads and a bounded queue"
> 
> Start in `start_metadata_update_broker()`. You'll see that depending on the rebalancer config it calls one of two functions:
> `metadata_update_broker_dynamic()`
> or
> `metadata_update_broker_dynamic()`
> 
Did you mean `metadata_update_worker_static()`?

agent/Cargo.toml
```diff
@@ -20,7 +20,7 @@ libmanta = { git = "https://github.com/joyent/rust-libmanta", tag = "v0.7.0" }
 mime = "0.3.13"
 moray = { git = "https://github.com/joyent/rust-moray", tag = "v0.9.2" }
 rebalancer = { path = "../rebalancer" }
-reqwest = "0.9.18"
+reqwest = "0.9.24"
```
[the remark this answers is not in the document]
rjloura: So this change can actually be backed out, or changed to `0.9` because cargo will automatically pick the highest patch version. 

Reason we aren't going to `0.10` is due to dependency conflicts with futures.  It would require a significant rewrite because many of the interfaces changed.

manager/src/jobs/status.rs
```diff
@@ -51,21 +63,27 @@ pub fn get_status(uuid: Uuid) -> Result<HashMap<String, usize>, StatusError> {
...
             Err(e) => {
                 error!("Status DB query: {}", e);
                 return Err(StatusError::LookupError);
             }
         };
 
-    for status_value in EvacuateObjectStatus::iter() {
-        let count = status_vec.iter().filter(|s| *s == &status_value).count();
-        let status_str = to_title_case(&status_value.to_string());
+    for status_count in status_counts.iter() {
```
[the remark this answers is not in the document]
rjloura: Yeah that's exactly what we do now.  Huge improvement on `GET /job` requests.  Certainly uses less memory.  What's weird is that the process should have reclaimed that memory either way, but it didnt.  In tests I've noticed that the little memory that is used by this approach is all reclaimed.

rjloura:
> > To review "MANTA-5160 rebalancer should allow for static number of threads and a bounded queue"
> > Start in `start_metadata_update_broker()`. You'll see that depending on the rebalancer config it calls one of two functions:
> > `metadata_update_broker_dynamic()`
> > or
> > `metadata_update_broker_dynamic()`
> 
> Did you mean `metadata_update_worker_static()`?

Yup.  My mistake.  I've updated that comment.

manager/src/main.rs
rjloura: done

manager/src/config.rs
rjloura: added some comments.  I think we will need to make a call on which knobs to expose via documentation soon.

manager/src/jobs/evacuate.rs
```diff
@@ -2730,6 +2818,96 @@ fn start_metadata_update_broker(
...
+fn metadata_update_broker_static(
+    job_action: Arc<EvacuateJob>,
+    md_update_rx: crossbeam::Receiver<AssignmentCacheEntry>,
+) -> Result<thread::JoinHandle<Result<(), Error>>, Error> {
+    thread::Builder::new()
+        .name(String::from("Metadata Update broker"))
+        .spawn(move || _update_broker_static(job_action, md_update_rx))
+        .map_err(Error::from)
+}
+
```
rhb2, line 2899: I need to verify my understanding of the big picture.  Help me out:

It looks to me like there are the two possible flows that can come from this:
```
metadata_update_broker_static
    _update_broker_static
        metadata_update_worker_static
             metadata_update_assignment
```
OR
```
metadata_update_broker_dynamic
    metadata_update_worker_dynamic
        metadata_update_assignment
```

In the case of the static path, the changes introduced appear to be:
1. It creates all threads that could be used for metadata updates up front with the understanding that some, or even most may have nothing to do, but there is a hard limit enforced by us on how many are created.
2. Because of (1), it also uses the condvar signal/wait so that these workers aren't busy spinning for no apparent reason.
3. The queue that is created in the static path is bounded unlike the one in the original (i.e. dynamic) path.

If I'm understanding the main things correctly about the static path, I like it.  I always like allocating all resources needed for a big operation up front so that if we make it past that point, we're likely home free.  I have to say though, it's not obvious to me why this will definitely save us memory, but even if it didn't, I still think it's a worthwhile addition.

In my opinion, at some point, if/when the static path has proven itself worthy, I hope we can get rid of the original path, but I guess right now, this is sort of a bake off between the two?  I definitely don't think that they should coexist forever and I hope we make plans to eventually commit to one.

Last but not least, I think it would be worth discussing in a very long, detailed comment above this function, the difference between the static and the dynamic approach and why they are both there for now.  I'd recommend leaving nothing to the imagination, especially if my understanding of it from above was way off.

rjloura: I am suspect that threadpools and unbounded queues are memory hogs.  The testing required to prove this has a fairly significant matrix and requires a significant amount of time.  In limited testing I have seen some anecdotal improvement, but the results are not consistent.

The dynamic approach allows us to tune the metadata tier impact while a job is running (we could probably build something similar with "static" threads).  I hate to completely remove that implementation because if it is not at fault for memory growth it makes live tuning much easier.  There's also the idea of an upper bound.

At this stage we have found a combination of changes that results in reduced memory consumption.  My hope is that we can put this back, get some solid testing on it, and confirm that we can evacuate over 50 million objects.  While testing I'd rather not have to generate new binaries to switch between these two approaches. But I generally agree that in the short term we should settle on one approach to rule them all.

There are some other issues around the bounded queue approach such as MANTA-5159.  Not an insurmountable issue, but something that probably wouldn't be an issue with an unbounded queue.

manager/src/jobs/evacuate.rs
```diff
@@ -638,6 +647,8 @@ impl EvacuateJob {
                 // TODO: should we panic? if so just replace with expect()
                 panic!("{}", e);
             });
+
+        self.remove_assignment_from_cache(assign_id);
```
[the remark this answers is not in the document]
rjloura: Correct.  We should now end the job with a completely empty hash.  This lowers our memory consumption because we dont bother to hold onto assignments that have been processed (skipped or otherwise).

manager/src/jobs/evacuate.rs
```diff
@@ -2377,6 +2390,8 @@ fn start_assignment_checker(
                         _ => continue,
                     }
 
+                    found_assignment_count += 1;
```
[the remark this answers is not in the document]
rjloura: I had intended to log this.  I just added one.

manager/src/jobs/evacuate.rs
```diff
@@ -589,6 +596,8 @@ impl EvacuateJob {
                 set_run_error(&mut ret, e);
             });
 
+        drop(job_action);
```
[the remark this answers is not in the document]
rjloura: I thought so too, but I wanted to try it to see if it made a difference.  As you know we have tried many different things here with varying results.  I removed it for now and will re-run some tests to see what happens.

manager/src/jobs/evacuate.rs
```diff
@@ -663,6 +674,10 @@ impl EvacuateJob {
             entry.is_some(),
             format!("Remove assignment not in hash: {}", assignment_id)
         );
+
+        drop(entry);
```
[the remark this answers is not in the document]
rjloura: Same comment as above.  I removed it for now and will re-run some tests to see what happens.

manager/src/jobs/evacuate.rs
```diff
@@ -663,6 +674,10 @@ impl EvacuateJob {
...
+
+        assignments.shrink_to_fit();
```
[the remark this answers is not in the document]
rjloura: I added some logging around this.

manager/src/jobs/evacuate.rs
```diff
@@ -2455,10 +2473,78 @@ fn reconnect_and_reupdate(
...
+                    if r.1.timed_out() {
+                        trace!("queue wait timed out");
+                    }
+                    continue;
+                }
+            };
+
+            let id = ace.id.clone();
+            trace!("Got assignment {}", id);
+            metadata_update_assignment(&job_action, ace, &mut client_hash);
```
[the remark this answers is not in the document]
rjloura: On the first iteration of the worker the `client_hash` is empty, but it is filled out with moray clients one the first and subsequent trip through `metadata_update_assignment()`.  This allows us to maintain the same clients for the duration of the worker threads, instead of closing all the connections, freeing the client and creating a new one each time. 

I shouldn't need to call `client_hash.clear()` in either one.  I suspect that those threads are not properly cleaning up the clients in the hash and possibly contributing to the memory leak, but since the `metadata_update_worker_static()` doesn't make that call and we are seeing much lower memory consumption that suspicion is probably unfounded.

manager/src/jobs/evacuate.rs
```diff
@@ -2481,154 +2567,155 @@ fn metadata_update_worker(
...
+                    &job_action,
+                    rmobj,
+                    shard,
+                    dest_shark,
+                    retag,
+                ) {
+                    Some((robj, rclient)) => {
+                        // Implicitly drop the client returned
+                        // from the hash which is probably bad
+                        client_hash.insert(shard, rclient);
```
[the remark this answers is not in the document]
rjloura: That would be good.  I am also thinking that we should remove this reconnect logic now that the associated cueball issues have been fixed.  I'll do that in a different wad though.

PR #73, opened 2020-04-14
MANTA-5122 - rebalancer-agent should run as nobody rather than root

ryanpuck:
Just a minor change in the smf for rebalancer-agent. This should run as nobody rather than root to avoid permission issues on target storage nodes

### DIFF
--- a/smf/manifests/rebalancer-agent.xml
+++ b/smf/manifests/rebalancer-agent.xml
@@ -7,7 +7,7 @@
 -->
 
 <!--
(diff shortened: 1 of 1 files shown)


ryanpuck:
This was tested and confirmed to work in qaperf1.

ryanpuck:
That's actually I'll I changed here however this smf was dumped from a running storage zone and I updated that dump rather than directly updating the smf from the repo. I'll go ahead and just update a copy of the one in master to lessen confusion. Thanks for bearing with me.

PR #76, opened 2020-04-20
MANTA-5196 rebalancer leaves some objects in assigned state

rjloura:


manager/src/jobs/evacuate.rs
```diff
@@ -830,6 +830,21 @@ impl EvacuateJob {
         Ok(ret)
     }
 
+    #[allow(clippy::ptr_arg)]
+    fn insert_assignment_into_db(
```
rhb2, line 834: Can we just have this function take a string and call it from everywhere?  I would have suggested variable arguments, but that can't be done as easily in rust as it is in C (I think it would require the creation of a macro like the ones that I created for logging).

manager/src/jobs/evacuate.rs
```diff
@@ -849,11 +862,8 @@ impl EvacuateJob {
                 panic!(msg);
             });
 
-        // TODO: remove.
-        // Leave this in here for now so that we can use the functionality
-        // quickly if we need to check something during development.
-        let mut total_time = self.total_db_time.lock().expect("DB time lock");
-        *total_time += now.elapsed().as_millis();
```
rhb2, line 856: We don't want to record this anymore?

manager/src/jobs/evacuate.rs
rjloura: I created this wrapper because I wanted to add some logging around when an assignment is inserted and when it is updated.

If you look closely you'll notice that we dont really insert an assignment, we actually insert a vector of `evacuateobject`s with a common `assignment_id`.

I was thinking that leaving `insert_many_into_db()` as a separate function would make sense for future use, but I could simply re-purpose that and remove the wrapper.

manager/src/jobs/evacuate.rs
rjloura: It was mostly useful before when we were using sqlite.  Now postgres is so much faster that it doesn't appear to have much of an impact.

manager/src/jobs/evacuate.rs
rhb2: Yes, I noticed that.  It calls `insert_many_into_db()`, and as far as I understood, that function doesn't take anything _but_ a vector of objects.  But that's actually my point.  The only reason this function appears to exist is to add a special logging message in the *only* instance we call it to insert the objects that comprise an assignment in to the database.  I didn't see any other place in the code where we call this function.  If it was to be called from several different sites, then I'd see it differently, but I don't know if it ever will be.  It seems like the same objective could be achieved with:
1. Creating a macro (which _will_ allow variable arguments if a custom message is what you want) and using that everywhere including the other places where you call `insert_many_into_db()`
2. If the macro is overkill because you don't feel like you need variable arguments (I'd get that), then would you consider a function that took a string (or even `Option<String>` for that matter) that everyone uses for inserting a vector of objects in to the database (regardless of whether the objects in the vector are tied to the same assignment or not)?
3. If both (1) and (2) are undesirable, then would it make sense to just add a single logging message above the call to `insert_many_into_db()` in this case?

I have a track record of creating utility functions that may only be called once just just to avoid implementing complicated logic in a deeply nested block.  It's just that in this case, the wrapper doesn't seem to exist for that reason.  It seems like it's just there to automatically log around a single use of `insert_many_into_db()` which could be accomplished with a single logging statement.  If we end up having to insert objects that comprise an assignment elsewhere in the manager then it would make more sense to me.  Are there plans in the foreseeable future to call this function anywhere else?

Honestly, my main concern is around confusing someone who knows a lot less about the manager than I do (and I don't think I set the bar high).  I can completely imagine someone not using one of these two things the right way, so why give them a choice?

manager/src/jobs/evacuate.rs
rhb2: Alright, I'll take your word for it.

manager/src/jobs/evacuate.rs
rjloura: Thats fine, I wasn't in love with that approach anyway.   I removed the indirection.

PR #83, opened 2020-04-28
MANTA-5212 Rebalancer agent: Incomplete objects should reside in a temporal location until complete

rhb2:


rhb2:
This took me a little longer than expected because I ran in to an issue while I was testing out the multithreaded agent.  Not that this change has any multithreading, but it does lay the groundwork for it so I decided to cheat and try it out.  It works really well.  Once this is integrated, the next change should be pretty straight forward.

For file rename operations and the like, if they we not successful, I elected to panic rather than create a new error code and return it for the manager to handle. I believe that if moving a file from one place to another fails, we would have big problems on our hands and we shouldn't be receiving any more assignments to process anyway.  That's where I was coming from at least.

rebalancer/src/libagent.rs
```diff
@@ -871,12 +882,29 @@ pub fn process_task(task: &mut Task) {
...
     // the object.
     let status =
         match download(&url, &task.owner, &task.object_id, &task.md5sum) {
-            Ok(_) => TaskStatus::Complete,
-            Err(e) => TaskStatus::Failed(e),
+            Ok(_) => {
+                // Upon successful download, move the temprorary object to its
+                // rightful location (i.e. /manta/account/object).
+                let manta_path = manta_file_path(&task.owner, &task.object_id);
+                if let Err(e) = fs::rename(&tmp_path, &manta_path) {
```
rjloura, line 895: Reading the documentation for `fs::rename()` I see this note:
> This will not work if the new name is on a different mount point.

I think /var/tmp and /manta are on two different mount points.  Have you tested this with a deployed manta?
```
[root@9f311b2f (storage) /var/tmp]$ df -h .
Filesystem                                  Size  Used Avail Use% Mounted on
zones/9f311b2f-041f-4207-ba2e-2488929de02b  259T  2.1G  259T   1% /
[root@9f311b2f (storage) /var/tmp]$ cd /manta/
[root@9f311b2f (storage) /manta]$ df -h .
Filesystem                                       Size  Used Avail Use% Mounted on
zones/9f311b2f-041f-4207-ba2e-2488929de02b/data  260T  1.1T  259T   1% /manta
```

rhb2: GOOD catch.  I read that in the documentation and completely forgot since /var/tmp to /manta don't cross a file system boundary in my development zone.  This definitely won't work.  Would you have any reservations about creating a subdirectory within /manta for this purpose?  It would almost always be empty anyway.

rjloura: I think that would be ok.

PR #85, opened 2020-04-29
MANTA-5221 Rebalanced agent: Increase the number of worker threads.

rhb2:


rebalancer/src/libagent.rs
```diff
@@ -899,24 +903,29 @@ pub fn process_task(task: &mut Task) {
...
-            }
-            Err(e) => {
-                // If we failed to complete the download, remove the temporary
-                // file so that these kinds of things do not pile up.  It is
-                // worth mentioning that in all failure cases except one there
-                // will a partially downloaded object that requires clean-up.
-                file_remove(&tmp_path);
-                TaskStatus::Failed(e)
-            }
-        };
```
rhb2, line 919: Cargo fmt strikes again.  No change to this code at all.

rebalancer/src/libagent.rs
```diff
@@ -1117,10 +1134,11 @@ pub fn router(f: fn(&mut Task), config: Option<AgentConfig>) -> Router {
 
         create_dir(REBALANCER_TEMP_DIR);
 
-        for _ in 0..1 {
+        for _ in 0..workers {
             let rx = Arc::clone(&rx);
             let assignments = Arc::clone(&agent.assignments);
             let m = agent_metrics.clone();
+            let client = reqwest::Client::new();
```
rhb2, line 1141: Here is where I give each worker their own client to work with.  It's wired all the way to do the download function where it's used to contact the mako.

rebalancer/src/libagent.rs
```diff
@@ -1090,20 +1100,27 @@ fn agent_start_metrics_server(config: &AgentConfig) -> MetricsMap {
 // Create a `Router`.  This function is public because it will have external
 // consumers, namely the rebalancer zone test framework.
 #[allow(clippy::many_single_char_names)]
-pub fn router(f: fn(&mut Task), config: Option<AgentConfig>) -> Router {
+pub fn router(
+    f: fn(&mut Task, &Client),
```
rjloura, line 1104: How is this going to jive with the always_pass function?  Are dummy functions going to have to create clients to satisfy this arg?

rebalancer/src/libagent.rs
```diff
@@ -1090,20 +1100,27 @@ fn agent_start_metrics_server(config: &AgentConfig) -> MetricsMap {
...
     build_simple_router(|route| {
         let mut agent_metrics: Option<MetricsMap> = None;
+        let mut workers = 1;
 
         if let Some(c) = config {
             agent_metrics = Some(agent_start_metrics_server(&c));
+            workers = c.server.workers;
         }
 
+        assert!(workers > 0);
```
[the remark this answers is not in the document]
rhb2: The comment about this always being true shocked me actually.  I've never heard of an unsigned value that couldn't at least be 0 (which is what it sounds like you're saying).  I had to see for myself, so I ran this:

```
    let lol = usize::min_value();
    println!("min size = {}", lol);
```
Sure enough, the output was:
```
min size = 0
```

Indeed, it can be 0.  I think it would be prudent to leave this assertion here -- if someone accidentally sets that tunable to 0, we want to know and quickly.  That was my rationale for having this check.

Regarding what would happen if we set it to -1, we will just fail to parse the config file due to a conflict in data type, so I'm not as worried about that.

During development though, I did unknowingly have the workers set to 0 and got a panic, but it was really really misleading because I didn't see it until gotham was handling a request and signaled one of the workers.  It was at that point that we panicked because there were no workers on the other end of the channel.

rebalancer/src/libagent.rs
rhb2: Turns out it was even easier than I thought while grocery shopping.  Change the signature of the function to take a `&Client` and simply do nothing with it.  It will be supplied to the function by `process_assignment()` regardless.  All the clients are created within the `router` function of the agent, so no need to worry about any of that.  Tests on the manager passed.

rebalancer/src/libagent.rs
```diff
@@ -37,7 +37,7 @@ use crate::metrics::{
     REQUEST_COUNT,
 };
 
-use reqwest::StatusCode;
+use reqwest::{Client, StatusCode};
```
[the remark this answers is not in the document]
rhb2: Will do.

rebalancer/src/libagent.rs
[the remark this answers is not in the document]
rjloura: clearly I was too tired to be doing this review... Obviously unsigned can be 0.  Next time I'll just go to bed :smile:

manager/src/jobs/evacuate.rs
```diff
@@ -3066,7 +3067,7 @@ mod tests {
     // declares the task as having been _successfully_ processed.  This will
     // generally be used when testing the "happy path" of most evacuation job
     // functionality.
-    fn process_task_always_pass(task: &mut Task) {
+    fn process_task_always_pass(task: &mut Task, client: &Client) {
```
rjloura, line 3070: Does this pass lint checking?  I would think you'd need to do something like:
```rs
fn process_task_always_pass(task: &mut Task, _client: &Client) {
```

rhb2: It'll pass cargo check only because it's test code which I think is not evaluated by the compiler at build time.  I saw the warning during test execution though.  I can take care of that.

PR #88, opened 2020-05-04
MANTA-5235 static metadata update queue is not necessary

rjloura:


manager/src/jobs/evacuate.rs
```diff
@@ -2936,26 +2918,31 @@ fn metadata_update_broker_dynamic(
...
+// The bounded channel will block on both send(if full) and receive(if empty).
 fn _update_broker_static(
     job_action: Arc<EvacuateJob>,
     md_update_rx: crossbeam::Receiver<AssignmentCacheEntry>,
 ) -> Result<(), Error> {
     let num_threads = job_action.options.max_metadata_update_threads;
     let queue_depth = job_action.options.static_queue_depth;
-    let queue = ArrayQueue::new(queue_depth);
-    let q_back = Arc::new(queue);
+    let (static_tx, static_rx) = crossbeam_channel::bounded(queue_depth);
```
rjloura, line 2935: This is the main part of the change. We are essentially using a crossbeam bounded channel as our "queue" here instead of implementing our own static queue with condition variables and sleeping when trying to push to a full queue.

PR #91, opened 2020-05-06
MANTA-5203 rebalancer should not automatically send assignments for every shark update

rjloura:
MANTA-5203 rebalancer should not automatically send assignments for every shark update
MANTA-5240 rebalancer should not allow old assignments to linger post MANTA-5203
MANTA-5239 rebalancer post and assignment generator threads could use better logging

### DIFF
--- a/manager/src/config.rs
+++ b/manager/src/config.rs
@@ -558,7 +558,9 @@ mod tests {
         File::open(TEST_CONFIG_FILE)
             .and_then(|mut f| {
                 let mut config_file = String::new();
+
                 f.read_to_string(&mut config_file).expect("config file");
+
                 assert!(config_file.contains("options"));
                 assert!(config_file.contains("max_tasks_per_assignment"));
                 assert!(config_file.contains("max_metadata_update_threads"));
@@ -566,6 +568,7 @@ mod tests {
                 assert!(config_file.contains("use_static_md_update_threads"));
                 assert!(config_file.contains("static_queue_depth"));
                 assert!(config_file.contains("max_assignment_age"));
+
                 Ok(())
             })
             .expect("config_basic_test");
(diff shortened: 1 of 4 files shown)


rhb2:
Looks to me like it was a bit more than a variable rename.  You're no longer hanging the maximum number of tasks allowed off of the job.  Instead, it's obtained from the config options which (if I'm understanding correctly hangs off of the evac job structure anyway, so you've cut down on retaining redundant information).  Let me know if I'm understanding the motivation for that correctly or not.  It may be worth just having a method on the evac job that can give us that information in the event that it moves around more in the future. :)

rjloura:
> Looks to me like it was a bit more than a variable rename. You're no longer hanging the maximum number of tasks allowed off of the job. Instead, it's obtained from the config options which (if I'm understanding correctly hangs off of the evac job structure anyway, so you've cut down on retaining redundant information). Let me know if I'm understanding the motivation for that correctly or not. It may be worth just having a method on the evac job that can give us that information in the event that it moves around more in the future. :)

I added "max_tasks_per_assignment" originally knowing that I wanted to eventually make it configurable.  When I did, I added "max_assignment_size" and totally forgot that I created "max_tasks_per_assignment".  

Essentially once I got around to actually using the variable I had renamed it, and the old version amounted to cuft.  It is really an oversight on my part.

rhb2:
On second thought, unless we can come up with a way of ensuring that this tunable does not get wiped from the config file when SAPI tunable changes, I think it might be a bit dangerous to introduce this knob -- or at least introduce it as something that can be consumed from the config file.

There are at least two options that I can think of off the top of my head:
* Throw it in the template file that config agent uses to generate the config
* Perhaps supply it as a command line argument when the service starts

There might be other ways beyond the scope of my knowledge, but those are the two primary mitigations that come to mind.  Either of these changes are pretty minor, so it probably wouldn't be a big deal to couple one with the introduction of this knob.

PR #95, opened 2020-05-14
MANTA-5233 rebalancer should use moray batch method for metadata updates

rjloura:


rebalancer/src/common.rs
```diff
@@ -312,3 +312,32 @@ pub fn get_objectId_from_value(
 
     Ok(id)
 }
+
+pub fn get_key_from_object_value(object: &Value) -> Result<String, Error> {
```
rjloura, line 316: This is a refactor from moray_client.rs

manager/src/jobs/evacuate.rs
```diff
@@ -2542,48 +2564,6 @@ fn start_assignment_checker(
         .map_err(Error::from)
 }
 
-fn reconnect_and_reupdate(
```
rjloura, line 2545: This was completely removed see: MANTA-5158

manager/src/jobs/evacuate.rs
```diff
@@ -2672,13 +2652,73 @@ fn metadata_update_worker_dynamic(
     }
 }
 
+fn get_client_from_hash<'a>(
```
rjloura, line 2655: This is a refactor from old line 2715

manager/src/jobs/evacuate.rs
```diff
@@ -2708,86 +2756,53 @@ fn metadata_update_assignment(
...
-                        // allow it to be added to the 'updated_objects'
-                        // Vec and later marked 'Complete' in the
-                        // database.
-                        robj
-                    }
-                    None => {
-                        continue;
-                    }
-                }
+                job_action.mark_object_error(&eobj.id, e.into());
```
rjloura, line 2805: This is where I coalesce all the errors from `update_object_shark()`

manager/src/jobs/evacuate.rs
```diff
@@ -1313,23 +1347,30 @@ impl GetAssignment for EvacuateJob {
 }
 
 impl UpdateMetadata for EvacuateJob {
-    /// This function only updates the local database in the event that an
-    /// error is encountered.  This allows the caller to easily batch database
-    /// updates for successful objects.
     fn update_object_shark(
```
rjloura, line 1350: The major changes to this function are:
1. remove the call to moray client to do the update
2. Coalesce error handling to the caller, which means removing the multiple calls `mark_object_error()`, and instead return an error code that can be easily converted to an EvacuateObjectError, and have that object inserted into the DB.

manager/src/jobs/evacuate.rs
```diff
@@ -2672,13 +2652,73 @@ fn metadata_update_worker_dynamic(
...
+        .map_err(|e| {
+            InternalError::new(
+                Some(InternalErrorCode::MetadataUpdateFailure),
+                e.description(),
+            )
+        })
+        .map_err(Error::from)
+}
+
 fn metadata_update_assignment(
```
rjloura, line 2709: I would recommend starting your review here.

manager/src/jobs/evacuate.rs
```diff
@@ -2708,86 +2756,53 @@ fn metadata_update_assignment(
...
-                        // Returning the retried object here will
-                        // allow it to be added to the 'updated_objects'
-                        // Vec and later marked 'Complete' in the
-                        // database.
-                        robj
-                    }
-                    None => {
-                        continue;
-                    }
-                }
```
rjloura, line 2790: This is removing all the reconnect / reupdate logic MANTA-5158

manager/src/moray_client.rs
```diff
@@ -128,30 +129,7 @@ pub fn put_object(
...
-        None => {
-            error!("Missing key field in object {:#?}", object);
-            return Err(InternalError::new(
-                Some(InternalErrorCode::BadMantaObject),
-                "Missing Manta Object Key",
-            )
-            .into());
-        }
-    };
+    let key = common::get_key_from_object_value(object)?;
```
rjloura, line 132: refactor

manager/src/jobs/evacuate.rs
```diff
@@ -2812,29 +2919,24 @@ fn metadata_update_assignment(
         updated_objects.push(eobj.id.clone());
```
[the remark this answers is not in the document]
rjloura: Pulled it out.

manager/src/jobs/evacuate.rs
rjloura: A lot of it is comments, formatting, and error handling (e.g. the call to `metadata_update_one()` takes 14 lines).  None the less I pulled some stuff out.

manager/src/jobs/evacuate.rs
```diff
@@ -2672,13 +2652,165 @@ fn metadata_update_worker_dynamic(
...
+}
+
+// Attempt to update all objects in this assignment in per shard batches.
+// So given an assignment with objects in two different shards, we will make
+// two calls to moray to update all the objects for this assignment in their
+// respective shards.
+//
+// If a batch fails this function falls back to updating each object
+// individually for that batch.
+fn metadata_update_batch(
```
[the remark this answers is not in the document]
rjloura: > I have some general questions about this function:
> 
>     1. Does the size of a batch make any difference at all when updating metadata?
> 

IIRC when I first put this wad together it did speed up metadata update times, but I am not sure how that scales.  i.e. perhaps bumping the batch from 100 to 1000 only saves us 1ms?

>     2. If the answer is yes, is there a limit to the batch size that is enforced anywhere?  I didn't see one in here.
> 

This is a good point.  Some brief research shows that there is no max PG transaction size.  Of course there are practical and resource limitations.  The limit right now is governed by the assignment size.  The code is written in such a way that a single batch update cannot have more than `max_tasks_per_assignment` put operations.  So it is bounded.

It probably makes sense to hard code some limits on all the tunables.

>     3. From the metadata tier standpoint, is there any difference between (for example) `metadata_upadate_one()` and just performing a batch update with a batch size of one (aside from the request name that is)?

I don't know if creating a PG transaction for a single operation is more or less expensive on the DB.  It does make things a little more complex for the rebalancer in that we have to build a batch of a single update instead of just passing that to `moray_client::put_object`.  What we gain in only using `mclient.batch()` we might lose in having to create a batch for a single update.  But, I think it is worth seeing what it looks like.  The only concern is that we have a fair amount of testing on the `moray_client::put_object()` approach, so it would be nice to have that as a fall back.

manager/src/jobs/evacuate.rs
```diff
@@ -2708,86 +2848,53 @@ fn metadata_update_assignment(
...
+                        key,
+                        value: o.clone(), // clone necessary?
+                    };
+
+                    batched_requests
+                        .entry(shard)
+                        .and_modify(|ent| {
+                            ent.push(BatchRequest::Put(put_req.clone()))
+                        })
+                        .or_insert_with(|| vec![BatchRequest::Put(put_req)]);
```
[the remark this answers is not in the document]
rjloura: Yes.  I refactored this piece.

rebalancer/src/error.rs
```diff
@@ -120,6 +120,8 @@ pub enum InternalErrorCode {
     SharkNotFound,
     DuplicateShark,
     BadMantaObject,
+    BadMorayClient,
+    MetadataUpdateFailure,
     JobBuilderError,
     MaxObjectsLimit,
 }
```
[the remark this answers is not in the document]
rjloura: I added some comments and also removed some.  Since the enum is `pub` the linter doesn't check for cruft.

manager/src/jobs/evacuate.rs
```diff
@@ -2672,13 +2681,209 @@ fn metadata_update_worker_dynamic(
...
+                        .expect("cannot get objectId");
+
+                    error!("Error updating object: {}\n{:?}", muo_err, o);
+                    job_action.mark_object_error(&id, muo_err.into());
+
+                    marked_error.push(id);
+
+                    continue;
+                }
+            }
```
rhb2, line 2844: Could we just have this be a separate function called (as an example, I'm not suggesting this specific name) `metadata_update_iterative()` which is essentially called from one of two places:
1. From `metadata_update_assignment()` if the config options specify that we aren't using batched updates.
2. From within `metadata_update_batch()` if we don't succeed in the batched update.

manager/src/jobs/evacuate.rs
rhb2: The changes that you've made so far have already helped make it more understandable (at least for me).  Thanks for that.

manager/src/jobs/evacuate.rs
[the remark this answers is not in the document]
rhb2: To be honest, my chief concern wasn't about how reliable the batch update functionality was, although if there are any reservations about that (aside from anyone just thinking that it needs more soak time), let me know.

Obviously, we're performing batched updates for a good reason.  One thing that has been gnawing at my subconscious is the fact that a single object update in the batch that fails has the potential to fully regress us back to the iterative/sequential update (or whatever we want to call it), putting us right back where we started.  The larger the batch size, the greater the risk.  That's good that the upper limit on batch size is the assignment length, but from a semantic standpoint, they really are two different things.  I am wondering if we should enforce a limit on batch size more explicitly, even if as of today, that value happens to be equal to the assignment size.  I have no idea what the right value would be, but at least in a case where we have one bad object update, we still end up retaining a significant amount of the benefit intended from this change.  An example of what I'm talking about would be if the batch size was half of the assignment size.  In a best case scenario, you make two requests instead of one, but if you have a single object failure, your worst case scenario is now only half as bad as what it would have been.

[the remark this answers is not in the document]
rjloura: I'd rather not expose additional knobs at this time.  In all of our testing we haven't seen a metadata update failure, so it is unlikely and adds additional complexity to an already complex change.  I'll open a jira task to keep an eye on this possibility, and if we find that it is common enough we can revisit.

manager/src/jobs/evacuate.rs
```diff
@@ -2708,133 +2913,66 @@ fn metadata_update_assignment(
...
+                if job_action.options.use_batched_updates {
+                    if let Err(e) =
+                        batch_add_putobj(&mut batched_reqs, o, shard, etag)
+                    {
                         error!(
-                            "MD Update Worker: failed to get moray \
-                             client for shard number {}. Cannot update \
-                             metadata for {:#?}\n{}",
-                            shard, mobj, e
+                            "Could add put object operation to batch ({}): {}",
```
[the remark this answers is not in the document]
rjloura: fixed

manager/src/jobs/evacuate.rs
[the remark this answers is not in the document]
rhb2: I don't necessarily think that it's a knob that needs to be publicly exposed.  You have introduced other like that already.  Like I said earlier though, as this is implemented, all it takes is one bad apple to negate all the benefits of this feature.  It seems rather unfortunate.

manager/src/jobs/evacuate.rs
rjloura: I refactored it.

manager/src/jobs/evacuate.rs
```diff
@@ -2672,13 +2687,228 @@ fn metadata_update_worker_dynamic(
...
+        })
+        .map_err(Error::from);
+
+    if ret.is_err() {
+        error!(
+            "Failed to update 1 object in {}us",
+            now.elapsed().as_micros()
+        );
+    } else {
+        error!("Updated 1 object in {}us", now.elapsed().as_micros());
```
rhb2, line 2757: How come this is expressed as an error?

manager/src/jobs/evacuate.rs
```diff
@@ -2672,13 +2687,228 @@ fn metadata_update_worker_dynamic(
...
+            error!("Error updating object: {}\n{:?}", muo_err, o);
+            job_action.mark_object_error(&id, muo_err.into());
+            marked_error.push(id);
+
+            continue;
+        }
+    }
+}
+
+fn batch_add_putobj(
```
rhb2, line 2875: Disclaimer: I readily acknowledge that I could be completely misunderstanding this part and if so, then I apologize in advance for the tedium.

If I understand correctly, the only way this can fail is if we are unable to get a key from the object passed in.  If this matters for adding objects to a batch for a batch update, would it also matter for performing singular updates?

If yes, then should we be checking for that somewhere in `update_assignment_metadata()` in the case where batch updates is not enabled?

If no, then in the case where we perform a retry of a batched update, would we want to retry all objects in the assignment as opposed to just the ones that we failed to initially add to the batch?

manager/src/jobs/evacuate.rs
rjloura: oops.  Fixed.

manager/src/jobs/evacuate.rs
rjloura: NP.  Good looking out.

If we fail to get the key for the batch put operation we fail that object and it is not included in the batch.

The call to `get_key_from_object_value()` is in the `moray_client::put_object()` method.

Either way it is only called once per object.

rhb2: Ok cool.  That makes sense.

rhb2:
I am approving this.  While I probably would have done a few things differently, I have to admit that the benefits of those things are subjective.  I think it's ready.

PR #96, opened 2020-05-14
MANTA-5246 rebalancer should be able to read object metadata from a file

rjloura:


manager/src/jobs/evacuate.rs
```diff
@@ -380,6 +722,97 @@ impl EvacuateObject {
             ..Default::default()
         }
     }
+
+    // This function converts a moray object to an evacuate object.
```
rjloura, line 726: This is a refactor from previous line number 1574.

PR #100, opened 2020-05-28
MANTA-5228 Rebalancer agent should process tasks in an assignment asynchronously

rhb2:


rebalancer/src/libagent.rs
```diff
@@ -1057,18 +1025,73 @@ fn process_assignment(
...
+    info!("Begin processing assignment {}.", &uuid);
+
+    let active_workers = min(len, pool.max_count());
+
+    for _ in 0..active_workers {
+        let a = Arc::clone(&assignment);
+        let fl = Arc::clone(&failures);
+        let id = uuid.clone();
+        let m = metrics.clone();
+        let c = client.clone();
```
rhb2, line 1084: Thought about making this a map of clients that are allocated once at the beginning.  I'm indifferent unless it turns out that this causes some kind of memory fragmentation.

rebalancer/src/libagent.rs
```diff
@@ -1057,18 +1025,73 @@ fn process_assignment(
...
+
+    info!("Begin processing assignment {}.", &uuid);
+
+    let active_workers = min(len, pool.max_count());
+
+    for _ in 0..active_workers {
+        let a = Arc::clone(&assignment);
+        let fl = Arc::clone(&failures);
+        let id = uuid.clone();
+        let m = metrics.clone();
```
rhb2, line 1083: This isn't cloning the actual metrics.  Those are global and modified atomically (prometheus does that).  For that reason, it didn't make sense (to me) to add an additional lock on this structure.  The extra memory usage that it temporarily incurs is marginal in the grand scheme of things and at any rate, it's reclaimed when the job(s) on the pool finish.

rebalancer/src/libagent.rs
```diff
@@ -1057,18 +1025,73 @@ fn process_assignment(
...
+        });
     }
+    pool.join();
 
-    let failed = if failures.is_empty() {
+    let failed = if failures.lock().unwrap().is_empty() {
         None
     } else {
-        Some(failures)
+        Some(failures.lock().unwrap().clone())
```
rhb2, line 1094: The way I did this was disappointing.  The failures vector is guarded by a mutex, but the thing that we want to assign it to is not.  AgentAssignmentState::Complete(Option<Vec<Task>>).  I didn't want to change that enum to have a mutex because it would have an impact on the manager too, so I resorted to the cheap trick of cloning the vector itself with the understanding that the original was about to be dropped.  That said, there is a short period of time where the two coexist.  It's not a big deal, but if memory is the concern, I can always play the game of popping elements off of the guarded vector and adding them to the unguarded vector until the contents have been fully transferred...unless you know of some more idiomatic way of doing it.

rebalancer/src/libagent.rs
rjloura: I believe the connection pooling is done at the Client level.  I am not sure if the pool is brought along with the clones.  If not then we are not taking full advantage of connection pooling.  Might be worth seeing if an Arc<Client> would work.  I am not sure it would.

Either way I am not sure it matters considering MANTA-5275.

rhb2: Thought about that too.  Even if we preserved them there is no guarantee that those connections will be relevant on the next assignment.  Somewhat related, my understanding was that different threads shouldn't share the same underlying client without serializing access.  Is that right?

rebalancer/src/libagent.rs
rjloura: I think this is fine.

rebalancer/src/libagent.rs
```diff
@@ -970,69 +974,31 @@ fn log_file_stats(task: &Task, uuid: &str) {
...
-        let mut t = {
-            let tmp = &mut assn.write().unwrap().tasks;
-            tmp[i].set_status(TaskStatus::Running);
-            tmp[i].clone()
+            tmp[index].set_status(TaskStatus::Running);
+            tmp[index].clone()
         };
 
         // Process it.
         f(&mut t, client);
```
rjloura, line 1000: I think what you have here is reasonable.  But I do have a question for clarity.

With the current approach you will iterate over all tasks n times (where n = min (tasks, <num_workers>)) taking the assignment write lock at least once for each iteration (twice for each task processed).

If you called `pool.execute()` here on each task you would only have to iterate over the assignment once 
```
pool.execute(move || {
  let result = f(&mut t, client);
  if let Some(res) = result {
    // take a lock on a failure vec
    // update the vec of failures
  }
});
```

This way you only iterate over the tasks once.  Another possible approach that would yield the same results would be to put the tasks in a queue and have a set of workers take work off that queue. 

Also, do we need to track the Pending and Running status of the tasks?  Or could we just set them as success or failed after the pool is joined and all workers are complete (based on the failures vec).  From the manager perspective we only really care if the assignment is scheduled, running, or complete (or failed/rejected for future use).  Then when it is complete we care about the status of the individual tasks.

rhb2: > I think what you have here is reasonable. But I do have a question for clarity.
> 
> With the current approach you will iterate over all tasks n times (where n = min (tasks, <num_workers>)) taking the assignment write lock at least once for each iteration (twice for each task processed).
> 
> If you called `pool.execute()` here on each task you would only have to iterate over the assignment once
> 
> ```
> pool.execute(move || {
>   let result = f(&mut t, client);
>   if let Some(res) = result {
>     // take a lock on a failure vec
>     // update the vec of failures
>   }
> });
> ```
I considered that approach.  The tradeoff being that if you have more tasks than there are workers in the pool, everything else will get queued once the workers are busy where the larger the assignment, the more that gets queued.  While each thread traverses through the list once, I highly doubt there's a performance impact given that downloading an object surely takes longer, but this subject is related to something else that you asked about (I'll tackle the tracking of assignment states below).  The other thing that I considered was removing the task from the list and then appending it to the end when I was done.  This would make access times for the next task to process constant and it would also avoid the queuing scenario that I mentioned, but there may be issues with that too.

> 
> This way you only iterate over the tasks once. Another possible approach that would yield the same results would be to put the tasks in a queue and have a set of workers take work off that queue.

Yeah, so this is what I had considered.  The problem with that is we really need to be performing work on the actual assignment that's been handed to us as opposed to doing work on a copy of it.  Why?  Because when the manager issues a request to get the latest assignment status, if we are operating on a copy of the assignment as opposed to the actual one, the manager won't see any progress at all until the whole thing is done.  I assumed that that was a non-starter.

> 
> Also, do we need to track the Pending and Running status of the tasks?
I track the state of the task while iterating through that loop because a status of `pending` is taken to be a task that has not been processed yet.  Anything other than `pending` is assumed to mean that it is either taken, or it's already done (for better or worse).

Or could we just set them as success or failed after the pool is joined and all workers are complete (based on the failures vec). From the manager perspective we only really care if the assignment is scheduled, running, or complete (or failed/rejected for future use). Then when it is complete we care about the status of the individual tasks.

rhb2: I think I have introduced something that will allow the best of both words.  Using a guarded index, workers can obtain the location of the next task to process without having to search.  Now there is no queueing (which I wasn't in favor of).

rhb2: With this tweak now, the number of times that we acquire a mutex should be something around 3*n where `n' is the length of the assignment:

1. To access a copy of and increment the index of the next task.
2. For updating the task in the assignment to "Running" and obtain a copy of it to process.
3. Updating results in the assignment at the completion of the task.

None of these operations take long, and I would be willing to wager that the majority of the time, when a thread goes for the lock it's not even blocked.  I should also throw out there that I'm not too worried about the cost of acquiring a mutex.  In comparison to everything else going on here, it's nothing, and the length of time for which its ever held is also *very* short.  You should always fence mutex acquisitions as tightly around the critical section as possible.  Further, I believe that the compare and swap associated with that kind of operation is performed in user land, so no system call.  Any performance gain that could be squeezed out further has to be darn near immeasurable and not worth the complexity of queuing on a thread pool or introducing additional auxiliary data structures.

rhb2: By the way, I don't see any reason why we can't just eliminate the `TaskStatus::Running` state altogether at this point. I believe that at the time it was introduced, we were initially giving back reports containing the status of all tasks as opposed to just specifics on ones that had failed, so it was a means of determining what object we were actually in the process of downloading.  If you don't think that the manager will ever have a need for that information, I don't think agent will either.  We can probably finally get rid of it.

rjloura: > By the way, I don't see any reason why we can't just eliminate the `TaskStatus::Running` state altogether at this point. I believe that at the time it was introduced, we were initially giving back reports containing the status of all tasks as opposed to just specifics on ones that had failed, so it was a means of determining what object we were actually in the process of downloading. If you don't think that the manager will ever have a need for that information, I don't think agent will either. We can probably finally get rid of it.

Correct.  I think this is a function of the transition from vision to implementation. 
I dont see a case where the manager would care about the state of individual tasks outside of their final result.  Even for future job types (if any) I assume we'd always operate on batches of objects (assignments).  What the manager ultimate needs to know is, the status of the overall assignment, and iff the assignment is complete, the result of each task.

I wonder if this would allow us to work off of a copy of the assignment...  i.e. take the assignment lock, set it to running, take a copy/clone, drop the lock.  This way other threads will take the lock, see the assignment is running, and ignore it and check the next assignment.  When the copied assignment completes it takes the lock again, updates the state to complete, and fills out the results (failures array etc), and sets the status to complete, then drops the lock.

rebalancer/src/libagent.rs
rjloura: I think you're right about that.  Even if we did try to use the same client across multiple threads I doubt it would buy us anything.

rebalancer/src/libagent.rs
rhb2: Huh?  I'm not really sure what you mean.  When we process an assignment, it's loaded directly from disk by a worker thread using nothing but a uuid that was communicated to it over a channel.  We aren't iterating over a bunch of assignments that are already loaded in memory and trying to find one to process.  Nothing is in memory until it's time to process it.  Maybe I misunderstood what you meant by that?  At any rate...

Three things:

The scope of this change was really only intended to take an existing process and parallelize it.  If we want to revisit the way we're fundamentally processing a task/assignment altogether _and_ it's going to be significant, lets go ahead and file another ticket for that so as not to introduce additional complexity to this one.  As it stands, we don't perform any additional locking on an assignment now than we did before this change and as an added bonus, we aren't using more memory / additional auxiliary data structures to do it.  The bonus was more important to me than extra locking if I'm being honest.

Entertaining some part of what I think you're getting at, even when processing a copy/clone of an assignment, it still has to be shared between multiple threads so locking is necessary, now just on the clone.  What do we get out of that?  Less contention with a client requesting the status of an assignment isn't a huge win given the frequency at which that happens, so hopefully there is something else.  Apologies if this question is irrelevant because I misunderstood what you're getting at.

Finally, and this is something that hasn't been discussed explicitly, but as a reminder, in the agent an assignment is actually more than a vector of tasks.  It's a vector of tasks + stats.  I think it's important to update some of the counters in those stats in real time so that it's possible to see progress on an assignment -- not just simply whether or not it's done.  That's fine if the manager doesn't need to know that we're halfway through an assignment right now, but that information is there so that if for some reason we do want to know how many more objects we have to go, an operator that knows what they're doing can answer that question.  I'm comfortable taking a poll on that one though if we're unsure.  That being said, if we're going to acquire the lock on an AgentAssignment in order to update stats, we may as well update the task while we're at it, especially since at the time of the update, the task is also complete.

rjloura: What you have is fine.  I thought that if we removed the requirement for per task status while the assignment was running it would make the code much simpler.  If you think it would introduce complexity then we shouldn't do it.

PR #101, opened 2020-06-01
MANTA-5289 deliver rebalancer agent as a separate image

timfoster:


timfoster:
This PR is the first part of the work to get a separate rebalancer agent image. We still need support in manta-adm to install this agent, and in manta-mako to drop the existing bundle agent.

agent-manifest.tmpl
```diff
@@ -0,0 +1,27 @@
...
+      "compression": "gzip"
+    }
+  ],
+  "description": "DESCRIPTION",
+  "tags": {
+    "smartdc_service": "true",
+    "buildstamp": "BUILDSTAMP"
+  },
+  "channels": [
+    "dev"
```
rjloura, line 25: I'm not very familiar with the manifest templates, so bear with me.  Does this mean that the image will only be available on the dev channel?

timfoster: Good question. I think this is boilerplate - it doesn't seem to impact where the image can be imported to. For example, the jenkins build for this work posted the following image to experimental:

https://updates.joyent.com/images/8fb8fa53-2490-4804-8fab-a5b1d2a3eee8?channel=experimental

which doesn't seem to have 'channels' at all. I'm looking around at the moment to see whether we can just drop this altogether.

PR #104, opened 2020-06-03
MANTA-5298 rebalancer min shard is always 1

rjloura:
I found the comment on line 102 while I was working on another change.  I figured I'd break this one out to make it easier to review.

### DIFF
--- a/manager/src/config.rs
+++ b/manager/src/config.rs
@@ -381,11 +381,11 @@ mod tests {
                         bld.insert_str("host", "200.fake.joyent.us")
                     })
                     .push_map(|bld| {
-                        bld.insert_str("host", "100.fake.joyent.us")
+                           bld.insert_str("host", "2.fake.joyent.us")
                     })
                     .push_map(|bld| {
                         bld
-                            .insert_str("host", "2.fake.joyent.us")
+                            .insert_str("host", "100.fake.joyent.us")
                             .insert_bool("last", true)
                     })
             }).build();
--- a/manager/src/config.rs
+++ b/manager/src/config.rs
@@ -99,29 +99,34 @@ pub struct Config {
 }
 
 impl Config {
-    // TODO: there's a bug here that 1 will always be the min shard number
     pub fn min_shard_num(&self) -> u32 {
-        self.shards.iter().fold(1, |res, elem| {
-            let shard_num = util::shard_host2num(elem.host.as_str());
+        let min_shard = self
+            .shards
+            .iter()
+            .min_by(|s, t| {
+                let s_num = util::shard_host2num(s.host.as_str());
+                let t_num = util::shard_host2num(t.host.as_str());
 
-            if res < shard_num {
-                return res;
-            }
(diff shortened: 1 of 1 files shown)


rjloura:
> I am wondering if we could sort the vector of shards at the time of creation with the intention of not having to iterate through them in order to find the smallest and largest.

We only go through this once at the beginning of a job.  So it is O(n) that one time in the worst case.  The shards come from SAPI. I dont think think the ordering is guaranteed there.

rhb2:
> > I am wondering if we could sort the vector of shards at the time of creation with the intention of not having to iterate through them in order to find the smallest and largest.
> 
> We only go through this once at the beginning of a job. So it is O(n) that one time in the worst case. The shards come from SAPI. I dont think think the ordering is guaranteed there.

Sure.  I expect that shard order is not guaranteed which is why I suggested sorting it.  Finding the min/max is O(n) for *every* job if I understand the code correctly, not just one job.  It looks to me like that information hangs off of a `config` structure which is passed in to the router function at the time that the service starts and that jobs created get a copy of that.  Over time, that turns in to 2 * m * n (not that the 2 technically matters) where `n' is the number of shards and `m' is the number of jobs.

If it's the case that every single job that gets created consults the config file directly such that every single job starts out with an unordered list of shards, then I'd say fine (maybe it is and I just missed it?).  Frankly, whether O(n) or O(n log n) or O(1), we all know that it's probably not an observable difference here anyway.  The spirit and intent of my suggestion was to make those two methods less complicated.

rhb2:
> The spirit and intent of my suggestion was to make those two methods less complicated.
Come to think of it (although I'm not asking for this), it might even make it possible for those methods to be non-existent if the vector was sorted...

rjloura:
> > The spirit and intent of my suggestion was to make those two methods less complicated.
> 
> Come to think of it (although I'm not asking for this), it might even make it possible for those methods to be non-existent if the vector was sorted...

My concern with this approach is that if someone creates a `Config{}` struct without calling `parse_config()` (or some other function that sorts the `Vec<Shard>`) then these functions will return incorrect values.  We have no real assurances that the `Vec<Shard>` will be sorted when that function is called.   We could re-work the Config module to have a ConfigBuilder and make the config structure private, but that would require a much more significant change that would effect many other modules.

Also consider that Jobs take on the order of hours to complete.  So having to iterate over 100ish shards twice every few hours won't make much of a difference on the aggregate.

rjloura:
I went ahead and made the shards field private.  This should prevent any other module from creating a `Config` without using `parse_config()` and prevent direct access to the shards Vec.

I also added some comments around the code to key in future developers to the fact that this Vec must remain sorted.

rhb2:
> I went ahead and made the shards field private. This should prevent any other module from creating a `Config` without using `parse_config()` and prevent direct access to the shards Vec.
> 
> I also added some comments around the code to key in future developers to the fact that this Vec must remain sorted.

Oh cool.  I will take a look.

PR #105, opened 2020-06-04
MANTA-5285 rebalancer could do a better job of tracking shark available MB

rjloura:
Posting this as a draft for now.  I still need to test it and make my own pass.


rjloura:
This still needs more work.

PR #108, opened 2020-06-12
MANTA-5303 Rebalancer Agent: Expose bytes/written to disk as a prometheus metric

rhb2:


rebalancer/src/libagent.rs
```diff
@@ -914,7 +914,16 @@ pub fn process_task(task: &mut Task, client: &Client) {
         &task.md5sum,
         client,
     ) {
-        Ok(_) => {
+        Ok(bytes) => {
+            if let Some(m) = metrics.clone() {
```
rjloura, line 918: I think the compiler might optimize this, but can we write it this way instead to avoid the clone?

```
if let Some(m) = metrics {
   counter_inc_by(m, BYTES_COUNT, bytes);
}
```

`metrics` is already an `&Option<MetricsMap>`

rebalancer/src/metrics.rs
```diff
@@ -97,6 +100,23 @@ pub fn counter_vec_inc_by<S: ::std::hash::BuildHasher>(
...
+    val: u64,
+) {
+    let num = val as f64;
+    match metrics.get(key) {
+        Some(metric) => {
+            if let Metrics::MetricsCounter(c) = metric {
+                c.inc_by(num);
+            }
+        }
+        None => error!(slog_scope::logger(), "Invalid metric: {}", key),
```
rjloura, line 116: if this is called by either the agent or the manager it should inherit the global scope and thus that of the logger, right?

rebalancer/src/libagent.rs
```diff
@@ -32,11 +32,7 @@ use joyent_rust_utils::file::calculate_md5;
 use libmanta::moray::MantaObjectShark;
 
 use crate::common::{AssignmentPayload, ObjectSkippedReason, Task, TaskStatus};
-use crate::metrics;
-use crate::metrics::{
-    counter_vec_inc, ConfigMetrics, MetricsMap, ERROR_COUNT, OBJECT_COUNT,
-    REQUEST_COUNT,
-};
+use crate::metrics::{self, *};
```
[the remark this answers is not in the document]
rhb2: Sure.

rebalancer/src/libagent.rs
```diff
@@ -842,7 +838,7 @@ fn download(
     object: &str,
     csum: &str,
     client: &Client,
-) -> Result<(), ObjectSkippedReason> {
+) -> Result<(u64), ObjectSkippedReason> {
```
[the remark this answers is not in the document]
rhb2: Done.

rebalancer/src/libagent.rs
rhb2: Fair enough.

rebalancer/src/metrics.rs
rhb2: `slog_scope::logger()` will access whatever `Logger` is in the current scope.  In both the case of the agent and the manager, there is only one and it's created when `init_global_logger()` is invoked.

rjloura: Doesn't the `error!()` macro already call `slog_scope::logger()` (in rebalancer::util)?  Or is there a different `error!()` macro in this scope?

rhb2: Ah.  Good question.  As you may recall, at the time that the metrics module was introduced, we weren't sure if we'd eventually want to pull this out in to another repository for other crates to use as well.  For that reason, we did not want it to depend on other constructs within the project if it didn't have to.  If you look a bit closer, you will see that every single logging message throughout metrics.rs follows the same convention -- it's not using the macros defined in `rebalancer::util`.  It's using the ones supplied to us by `slog`.  See near the top of the file (I think maybe line 21?).

At any rate, since this probably won't ever be consumed by anything else at this point, we could use the macros defined in util.  I know what it'd take to do it and it'd involve touching every logging statement in metrics.rs along with a change to util.rs.  If we want to go that route, mind if we do that as part of a separate ticket?

rjloura: Ah I see.  I think we should leave it as is.

rhb2: Appreciate you taking the time.

PR #112, opened 2020-06-19
MANTA-5324 Introduce histogram metric for representing assignment processing times

rhb2:


rebalancer/src/libagent.rs
```diff
@@ -33,8 +33,9 @@ use libmanta::moray::MantaObjectShark;
 
 use crate::common::{AssignmentPayload, ObjectSkippedReason, Task, TaskStatus};
 use crate::metrics::{
-    self, counter_inc_by, counter_vec_inc, ConfigMetrics, MetricsMap,
-    BYTES_COUNT, ERROR_COUNT, OBJECT_COUNT, REQUEST_COUNT,
+    self, counter_inc_by, counter_vec_inc, histogram_observe, ConfigMetrics,
```
bowrocker, line 36: This list is getting a little lengthy; maybe consider changing to:

`use crate::metrics::{ self, *};`

rhb2: Sounds good, Jon.  Done!

PR #114, opened 2020-06-23
MANTA-5328 metadata update throttle does not immediately stop extra running threads

rjloura:


manager/src/jobs/evacuate.rs
```diff
@@ -3045,13 +3045,21 @@ fn metadata_update_worker_dynamic(
...
                 },
                 Steal::Retry => continue,
                 Steal::Empty => break,
             };
             metadata_update_assignment(&job_action, ace, &mut client_hash);
         }
+
+        debug!(
+            "Exiting metadata update worker: {:?}",
+            thread::current().id()
```
rhb2, line 3061: The logging macros in rebalancer/src/util.rs will print the thread name for you without having to explicitly provide it here.  If it's the case that the thread is named, then maybe you won't actually see the thread id and you'll see something else instead?

At any rate, we're already using https://crates.io/crates/thread-id elsewhere in the rebalancer which will allow you to obtain the numeric value for the thread id as opposed to something that looks like `ThreadId(13)` for example.

manager/src/jobs/evacuate.rs
```diff
@@ -3392,8 +3400,17 @@ fn update_dynamic_metadata_threads(
...
+    let difference: i32 = new_worker_count as i32 - *max_thread_count as i32;
+
+    debug!(
+        "Updating metadata update thread count to {}.",
+        new_worker_count
+    );
+
+    if difference < 0 {
+        debug!("Need to stop {} threads", difference.abs());
+    }
```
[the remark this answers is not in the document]
rjloura: good point.

manager/src/jobs/evacuate.rs
rjloura: good point.

PR #117, opened 2020-06-25
MANTA-5216 rebalancer log should default to debug level

rjloura:


rjloura:
> The change makes sense to me, especially since we know how expensive the verbose logging can be. The only thing that I'm kind of on the fence on is whether or not a config file is the best way to control it. Have you considered attempting to obtain it from an environment variable instead?

I did, and I've seen that before, but from what I've seen the convention in triton/manta is to use a config file (cnapi, napi, muskie, moray, etc).  I don't see much of a benefit either way, other than with a config file all the startup args are in the same place, a one stop shop so to speak.

rhb2:
> > The change makes sense to me, especially since we know how expensive the verbose logging can be. The only thing that I'm kind of on the fence on is whether or not a config file is the best way to control it. Have you considered attempting to obtain it from an environment variable instead?
> 
> I did, and I've seen that before, but from what I've seen the convention in triton/manta is to use a config file (cnapi, napi, muskie, moray, etc). I don't see much of a benefit either way, other than with a config file all the startup args are in the same place, a one stop shop so to speak.

While I'm not necessarily a fan of controlling it that way, doing my due diligence, I reviewed some of the more prominent components in Manta and indeed, they control the logging level via a config file as well.  One advantage of doing it that way I guess would be that if you have (for example) 1000 instances of a service running and you see that they are all generating an obscene amount of logs (that you aren't currently interested in), you can dial them all back at once very quickly.  That's probably the only advantage that I see to doing it this way over the other.  And sadly, it doesn't even look like most components take advantage of that!  Either way, you are following convention so for that reason I'll stamp this.

rjloura:
> If you haven't already tested the agent tests, don't forget to give that a run too.

Agent unit tests looked good.  I need to do some tests on my dev env still.

rhb2:
> > If you haven't already tested the agent tests, don't forget to give that a run too.
> 
> Agent unit tests looked good. I need to do some tests on my dev env still.

Cool.  I forgot that this was still in a draft state.  If/when another stamp is needed after making it official, let me know.

manager/src/main.rs
```diff
@@ -567,14 +558,23 @@ fn main() {
...
-    let config = Arc::new(Mutex::new(
-        Config::parse_config(&config_file)
-            .map_err(|e| {
-                error!("Error parsing config: {}", e);
-                std::process::exit(1);
-            })
-            .unwrap(),
-    ));
+    let config = Config::parse_config(&config_file)
+        .map_err(|e| {
```
kellymclaughlin, line 562: You could shorten this up just a bit by using `unwrap_or_else` instead of `map_err` and `unwrap`.

sapi_manifests/rebalancer/template
```diff
@@ -9,6 +9,7 @@
         "use_batched_updates": true
     },
     "listen_port": 80,
+    "log_level": "debug",
```
kellymclaughlin, line 12: Should this be configurable via a SAPI variable as well?

rjloura: why not!

manager/src/main.rs
rjloura: done.

PR #118, opened 2020-06-26
MANTA-5287 determine final list of rebalancer tunables and make them SAPI aware

rjloura:


sapi_manifests/rebalancer/template
```diff
@@ -1,12 +1,56 @@
...
         "max_assignment_age": 600,
+        {{/REBALANCER_MAX_ASSIGNMENT_AGE}}
+
+
+        {{#REBALANCER_USE_BATCHED_UPDATES}}
         "use_batched_updates": true
+        {{/REBALANCER_USE_BATCHED_UPDATES}}
+        {{^REBALANCER_USE_BATCHED_UPDATES}}
+        "use_batched_updates": false
+        {{/REBALANCER_USE_BATCHED_UPDATES}}
```
rhb2, line 53: I think this basically says that if this tunable exists at all, regardless of its value, then we set `use_batched_updates` to `true`, otherwise, if it's not present, set it to false?  Am I understanding the intention correctly?

I was expecting to see something more like this:

```
{{#REBALANCER_USE_BATCHED_UPDATES}}
"use_batched_updates": {{REBALANCER_USE_BATCHED_UPDATES}}
{{/REBALANCER_USE_BATCHED_UPDATES}}
{{^REBALANCER_USE_BATCHED_UPDATES}}
"use_batched_updates": false
{{/REBALANCER_USE_BATCHED_UPDATES}}
```

I saw at least one other place where you did this with another boolean tunable as well further up.

rhb2: Despite the fact that it's a boolean, I think we should use a value explicitly to set the state of this latch.

sapi_manifests/rebalancer/template
```diff
@@ -1,12 +1,56 @@
...
+
+        {{#REBALANCER_MAX_SHARKS}}
+        "max_sharks": {{REBALANCER_MAX_SHARKS}},
+        {{/REBALANCER_MAX_SHARKS}}
+        {{^REBALANCER_MAX_SHARKS}}
         "max_sharks": 5,
+        {{/REBALANCER_MAX_SHARKS}}
+
+        {{#REBALANCER_USE_STATIC_MD_UPDATE_THREADS}}
+        "use_static_md_update_threads": true,
```
kellymclaughlin, line 27: So the idea here I guess is that the SAPI variable acts as a flag and this is just set to true regardless of the SAPI variable value?

rjloura: yes, but I'm thinking it would be better to swap that to be like `REBALANCER_USE_BATCHED_UPDATES` with a default of false.

kellymclaughlin: :+1:

sapi_manifests/rebalancer/template
rjloura: Its a good point.  Updated.

PR #119, opened 2020-06-30
MANTA-5332 rebalancer: want timer that ignores initial idle scan time

rjloura:


rjloura:
> Looks fine. I'd consider just making that info!() to be honest since I don't think it'll cost us too much in the way of logging, but I'm fine with whatever you decide.

That's a good idea.  I changed the other log messages as well.

PR #121, opened 2020-07-01
MANTA-5334 timer for MANTA-5332 could be improved for max objects and multiple shards

rjloura:


manager/src/jobs/evacuate.rs
```diff
@@ -2129,6 +2120,21 @@ where
...
+                            .unwrap();
+
+                        if let Some(st) = *start_time {
+                            info!(
+                                "Evacuate Job object movement time: {} seconds",
+                                st.elapsed().as_secs()
+                            );
+                        }
+
+                        *start_time = None;
```
papertigers, line 2136:
```Rust
  2130                         if let Some(st) = start_time.take() {
  2131                             info!(
  2132                                 "Evacuate Job object movement time: {} seconds",
  2133                                 st.elapsed().as_secs()
  2134                             );
  2135                         }
```

I think something like this would be cleaner and allows you to ditch the `*start_time = None` bit.

bahamas10:
:+1: I pulled the code and looked through it... this change makes sense.  I also looked into `.take()` as @papertigers mentioned and it seems like that may be a slightly cleaner/optimal approach.

manager/src/jobs/evacuate.rs
rjloura: Good tip!

PR #123, opened 2020-07-08
MANTA-5339 rebalancer: job endpoint should expose job parameters

rjloura:


manager/src/jobs/evacuate.rs
```diff
@@ -254,11 +269,67 @@ enum MetadataClientOption<'a> {
...
+    create_query: &str,
 ) -> Result<usize, Error> {
-    let status_strings = EvacuateObjectStatus::variants();
-    let error_strings = EvacuateObjectError::variants();
+    // TODO: check if table exists first and if so issue warning.  We may
+    // need to handle this a bit more gracefully in the future for
+    // restarting jobs.
+    let drop_query = format!("DROP TABLE {}", table_name);
+
+    conn.execute(&drop_query).unwrap_or_else(|e| {
```
papertigers, line 282: I think it's more idiomatic to do `if let Err(e) = foo() { ... }`

manager/src/jobs/evacuate.rs
```diff
@@ -254,11 +269,67 @@ enum MetadataClientOption<'a> {
...
+        from_shark: from_shark_value,
+    };
+
+    let updated_records = diesel::insert_into(config_table)
+        .values(&value)
+        .on_conflict(config_id)
+        .do_update()
+        .set(&value)
+        .execute(conn)
+        .unwrap();
```
papertigers, line 319: Comment why this unwrap is safe?

manager/src/jobs/evacuate.rs
```diff
@@ -310,7 +382,8 @@ pub fn create_evacuateobjects_table(
             );",
         status_check, skipped_check, error_check
     );
-    conn.execute(&create_query).map_err(Error::from)?;
+
+    create_table_common(conn, "evacuateobjects", &create_query)?;
```
papertigers, line 386: Is it worth making "evacuateobjects" a const that can be used elsewhere rather than typing out the string every time?

manager/src/jobs/status.rs
```diff
@@ -131,20 +241,27 @@ mod tests {
...
         let mut obj_vec = vec![];
-        let conn = pg_db::create_and_connect_db(&uuid.to_string()).unwrap();
+        let config = Config::default();
+        let job_builder = JobBuilder::new(config);
 
-        evacuate::create_evacuateobjects_table(&conn).unwrap();
+        let job = job_builder
+            .evacuate("fake_shark".to_string(), Some(NUM_OBJS as u32))
+            .commit()
+            .unwrap();
```
papertigers, line 260: Again, is this safe?

manager/src/jobs/status.rs
```diff
@@ -131,20 +241,27 @@ mod tests {
...
+
+        let job_id = job.get_id();
+
+        let conn = pg_db::connect_or_create_db(&job_id.to_string()).unwrap();
```
papertigers, line 264: Unwrap agian

manager/src/jobs/status.rs
```diff
@@ -155,9 +272,11 @@ mod tests {
             .execute(&conn)
             .unwrap();
 
-        let count = get_status(uuid.clone()).unwrap();
+        let job_status = get_job(job_id).unwrap();
```
papertigers, line 275: Unwrap safe?

manager/src/jobs/status.rs
```diff
@@ -155,9 +272,11 @@ mod tests {
...
+        let JobStatusResults::Evacuate(evac_job_results) = job_status.results;
+        let count = *evac_job_results.get("Total").unwrap();
```
papertigers, line 277: Unwrap safe?

manager/src/jobs/status.rs
```diff
@@ -166,12 +285,18 @@ mod tests {
...
         let mut obj_vec = vec![];
-        let conn = pg_db::create_and_connect_db(&uuid.to_string()).unwrap();
 
-        evacuate::create_evacuateobjects_table(&conn).unwrap();
+        let config = Config::default();
+        let job_builder = JobBuilder::new(config);
+        let job = job_builder
+            .evacuate("fake_shark".to_string(), Some(NUM_OBJS as u32))
+            .commit()
+            .unwrap();
```
papertigers, line 296: Unwrap safe? (Feel free to dismiss these all if they are all safe to unwrap)

manager/src/jobs/status.rs
```diff
@@ -166,12 +285,18 @@ mod tests {
...
+
+        let job_id = job.get_id();
+        let conn = pg_db::connect_db(&job_id.to_string()).unwrap();
```
papertigers, line 299: Unwrap safe?

manager/src/jobs/status.rs
```diff
@@ -186,10 +311,13 @@ mod tests {
             .execute(&conn)
             .unwrap();
 
-        let count = get_status(uuid.clone()).unwrap();
+        let job_status = get_job(job_id).unwrap();
```
papertigers, line 314: Unwrap safe? ( I will stop leaving these now :stuck_out_tongue: )

manager/src/jobs/evacuate.rs
rjloura: used map_err()

manager/src/jobs/evacuate.rs
rjloura: Sure, one less line.

manager/src/jobs/evacuate.rs
rjloura: its only used here, but your point is none the less valid.  Changed.

manager/src/jobs/status.rs
rjloura: This is test code, but I changed to expect() to get more info.

manager/src/jobs/status.rs
rjloura: test code => expect()

manager/src/jobs/status.rs
rjloura: test code => expect()

manager/src/jobs/status.rs
rjloura: test code => expect()

manager/src/jobs/status.rs
rjloura: test code => expect()

manager/src/jobs/status.rs
rjloura: test code => expect()

manager/src/jobs/status.rs
rjloura: test code => expect() :)

rjloura:
> These changes look fine to me, but the jobs tests don't pass for me. There's apparently some database setup that's not part of the test that I'm missing. Before I approve can you confirm they pass for you?
> 
> ```
> failures:
> 
> ---- jobs::status::tests::get_status_test stdout ----
> thread 'main' panicked at 'job builder: Diesel(DatabaseError(__Unknown, "relation \"jobs\" does not exist"))', src/libcore/result.rs:1165:5
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
> 
> ---- jobs::status::tests::get_status_zero_value_test stdout ----
> thread 'main' panicked at 'job builder: Diesel(DatabaseError(__Unknown, "relation \"jobs\" does not exist"))', src/libcore/result.rs:1165:5
> 
> ---- jobs::status::tests::list_job_test stdout ----
> thread 'main' panicked at 'slog-scope: No logger set. Use `slog_scope::set_global_logger` or `slog_scope::scope`.', /home/kelly/.cargo/registry/src/github.com-1ecc6299db9ec823/slog-scope-4.3.0/lib.rs:125:13
> 
> ---- jobs::test::basic stdout ----
> thread 'main' panicked at 'failed to create job: Diesel(DatabaseError(__Unknown, "relation \"jobs\" does not exist"))', src/libcore/result.rs:1165:5
> 
> 
> failures:
>     jobs::status::tests::get_status_test
>     jobs::status::tests::get_status_zero_value_test
>     jobs::status::tests::list_job_test
>     jobs::test::basic
> 
> test result: FAILED. 10 passed; 4 failed; 0 ignored; 0 measured; 5 filtered out
> 
> error: test failed, to rerun pass '-p manager --lib'
> ```

I think that one is due to the fact that we dont bother to migrate the existing databases, and the tests don't do a great job of cleaning up after themselves.  So when you run a test it will probably pick up an older database with and older schema.  Since this is the first production release I didn't bother to build in migrate code.

I did however notice a different test issue which I am going to push a fix for now.

manager/src/jobs/evacuate.rs
```diff
@@ -616,8 +704,15 @@ impl EvacuateJob {
         Ok(())
     }
 
+    fn update_evacuate_config(&self) -> Result<usize, Error> {
+        let locked_conn = self.conn.lock().expect("DB conn lock");
+
+        update_evacuate_config_impl(&*locked_conn, &self.from_shark)
```
papertigers, line 710: I think you can ditch the `*` here. At least rust 1.41 seems happy with my change.
One recommendation is that we add a cargo override file for our projects that match the rust version we are using with jenkins.  This way anyone looking at the repo in osx/linux end up using the same rust version.

manager/src/jobs/status.rs
papertigers: Cool, works for me.  I think more recent versions or rust will actually give you line number etc on unwrap() now.  But still nice to use expect.

PR #127, opened 2020-07-21
MANTA-5353 rebalancer: make sharkspotter chunksize tunable

rjloura:


manager/src/config.rs
```diff
@@ -52,6 +52,9 @@ static DEFAULT_STATIC_QUEUE_DEPTH: usize = 10;
 // destination sharks.
 static DEFAULT_MAX_ASSIGNMENT_AGE: u64 = 600;
 
+// The chunk sized used when scanning the metadata tier.
```
rhb2, line 55: 's/sized/size'

PR #128, opened 2020-07-21
MANTA-5356 Rebalancer Manager: Should have metrics port accessible on admin and manta IP

rhb2:


rebalancer/src/metrics.rs
```diff
@@ -42,7 +42,7 @@ pub struct ConfigMetrics {
 impl Default for ConfigMetrics {
     fn default() -> Self {
         Self {
-            host: "127.0.0.1".into(),
+            host: Ipv4Addr::UNSPECIFIED.to_string(),
```
rhb2, line 45: https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#associatedconstant.UNSPECIFIED

PR #129, opened 2020-07-24
MANTA-5246 rebalancer evacuate job should be able to read objectIds from a file

rjloura:
Still some work to do, but posting as a Draft PR for now.


manager/src/jobs/evacuate.rs
```diff
@@ -405,6 +418,630 @@ pub fn create_evacuateobjects_table(
...
+
+    // XXX check shard number !> i32::MAX
+    eobj.shard = shard_num;
+
+    Ok(eobj)
+}
+
+fn walk_obj_ids_in_file(
+    job_action: Arc<EvacuateJob>,
+    shard_file: PathBuf,
```
papertigers, line 669: You could make this more flexible and do `<P: AsRef<Path>>` and then calling functions could just pass something like a String/str directly without having to do PathBuf::from().

manager/src/jobs/evacuate.rs
```diff
@@ -405,6 +418,630 @@ pub fn create_evacuateobjects_table(
...
+            min_shard: self.min_shard,
+            max_shard: self.max_shard,
+            sharks: vec![shark.to_string()],
+            chunk_size: job_action.config.options.md_read_chunk_size as u64,
+            ..Default::default()
+        };
+
+        debug!("Starting sharkspotter generator: {:?}", &config);
+
+        let log = slog_scope::logger();
```
rhb2, line 975: Looks to me like the only thing this variable is used for is to pass it to `sharkspotter::run_multithreaded()`.  You could probably just pass `slog_scope::logger()` as an argument instead on line 991, but that's just a nit and an opinion.

manager/src/rebalancer-adm.rs
```diff
@@ -161,8 +175,6 @@ fn main() -> Result<(), String> {
                         .about("Get information on a specific job")
                         .arg(
                             Arg::with_name("uuid")
-                                .short("u")
-                                .long("uuid")
```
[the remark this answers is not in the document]
rjloura: IIRC there was some grumbling from ops/QA.

manager/src/rebalancer-adm.rs
```diff
@@ -139,6 +146,13 @@ fn main() -> Result<(), String> {
                 .required(true)
                 .help("Specifies a shark on which to run a job"),
         )
+        .arg(
+            Arg::with_name("file_source")
+                .short("F")
+                .long("file_source")
+                .takes_value(true)
```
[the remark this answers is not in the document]
rjloura: The file(s) will need to be accessible via a file path on the host running rebalancer.  Granted `rebalancer-adm` could be run from another location, but the URL for interfacing with the manager is still hard coded to `localhost`.  Until that changes attempts at running `rebalancer-adm` remotely would fail.  

It may have been a better approach to have the client stream object IDs over to the rebalancer server instead of having the server read them in from a file.  If I had more time I'd probably do that now.  There are some other considerations with that approach such as how concurrent the per-shard file reads (and by implication object ID stream) should be.  It also creates a new set of failure modes, etc. 

I went this direction because I was thinking long term we may want to have a `scan` job where the operator would initiate a scan with `rebalancer-adm`.  That scan would either output the object ids into a set of files on the server for use later.  For this long term approach it makes more sense to start with having the files on the server as opposed to where the client is located (which is the same place for now).  Thinking even longer term we could have a job that both scans for multiple sharks, and evacuates one of them concurrently, while saving off the object IDs into files which it will automatically start reading back in once the first shark has completed evacuated.

To be clear, the current approach requires the operator to run sharkspotter on a host that has a connection to the manta network, and then place those files in a path that is accessible by the rebalancer manager/server.

manager/src/jobs/evacuate.rs
```diff
@@ -405,6 +418,630 @@ pub fn create_evacuateobjects_table(
...
+        let reader = BufReader::new(File::open(shard_file)?);
+        let msg = format!("Could not find object with id: {}", object_id);
+
+        for json_obj in reader.lines() {
+            let json_obj = json_obj?;
+
+            let moray_object: MorayObject =
+                serde_json::from_str(json_obj.as_str())?;
+
+            let manta_object = moray_object.value.clone();
```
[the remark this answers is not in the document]
rjloura: Done.

manager/src/rebalancer-adm.rs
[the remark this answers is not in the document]
rhb2: > To be clear, the current approach requires the operator to run sharkspotter on a host that has a connection to the manta network, and then place those files in a path that is accessible by the rebalancer manager/server.

> This is really a stepping stone that, while requiring a fair amount of operator intervention, should quickly and significantly alleviate the performance issues due to sparse objects across the MD tier.

Yeah, this is kind of what I figured too and especially given the little time that we have, I would have done it the exact same way.  Something along the lines of those last two sentences are perfect for a copy/paste in to one of the README files.

manager/src/jobs/evacuate.rs
rjloura: Ok.  As of now it is only called with a PathBuf, but I can see the future utility.

manager/src/jobs/evacuate.rs
rjloura: :+1:

manager/src/jobs/evacuate.rs
```diff
@@ -405,6 +418,634 @@ pub fn create_evacuateobjects_table(
...
+    }
+
+    fn walk_shards(&self) -> Result<(), Error> {
+        let max_threads =
+            self.job_action.config.options.max_metadata_read_threads;
+        let pool = threadpool::ThreadPool::with_name(
+            "File Generator".into(),
+            max_threads,
+        );
+
```
rjloura, line 864: comment here

manager/src/jobs/evacuate.rs
```diff
@@ -405,6 +419,645 @@ pub fn create_evacuateobjects_table(
...
+    fn get_object(
+        &mut self,
+        object_id: &ObjectId,
+    ) -> Result<MorayObject, Error> {
+        let shard_file = self.shard_file.clone();
+        let reader = BufReader::new(File::open(shard_file)?);
+        let msg: String;
+
+        for json_obj in reader.lines() {
+            let json_obj = json_obj?;
```
rhb2, line 502: This line is kind of confusing to me.  Is this intended to be variable shadowing (or whatever it's called in the documentation)?

manager/src/jobs/evacuate.rs
```diff
@@ -405,6 +419,645 @@ pub fn create_evacuateobjects_table(
...
+        &mut self,
+        object_id: &ObjectId,
+    ) -> Result<MorayObject, Error>;
+}
+
+struct FileObjectGetter {
+    shard_file: PathBuf,
+}
+
+impl ObjectGetter for FileObjectGetter {
```
rhb2, line 492: In my opinion this function is one of the most interesting parts of the change.  The walkthrough yesterday cleared up a lot of confusion about what's going on.  For the uninitiated though, I think they might confuse what's happening here with what's happening in real life (i.e. in production, take an object id from a file and resolve through moray, vs take an object id from a file resolve through another file).  I think most people won't realize that there are two types of files that serve two different purposes.  I guess that's the long way of saying that I think some commentary clarifying that,  either here or somewhere where you think it's more likely to be seen, would be good.

manager/src/jobs/evacuate.rs
```diff
@@ -405,6 +419,645 @@ pub fn create_evacuateobjects_table(
...
+            "File Generator".into(),
+            max_threads,
+        );
+
+        // This call gets the top level directory which we iterate over
+        // below.  Each thread will get a path like:
+        // `/user/specified/<shark name>/shard_#.objs`
+        let dir = self.get_source_dir().map_err(|e| {
+            InternalError::new(
+                Some(InternalErrorCode::InvalidJobAction),
```
rhb2, line 877: How come we categorize this as an `InvalidJobAction`?  Looks like it can fail primarily if `path_buf_from_dirname()` fails which may be where my understanding falls short.  I went over that function really quick, but for the most part, it looks like it will succeed if a decent path was given that meets the criteria enforced in that function.

manager/src/jobs/evacuate.rs
rjloura: `reader.lines()` returns a iterator of a `io::Result<String>`, so I need to handle that result and get the inner string.  Something like this won't work because `in` assumes an iterator not a Result:
```
for json_obj in reader.lines()? { }
```

I'm open to a more idiomatic way of doing this.

manager/src/jobs/evacuate.rs
rjloura: Maybe calling this `InvalidJobParam` would make more sense.  The idea is that the user provided a path that was not valid or usable for this job.

manager/src/jobs/evacuate.rs
rjloura: comment added.

manager/src/jobs/evacuate.rs
```diff
@@ -405,6 +419,663 @@ pub fn create_evacuateobjects_table(
 
 struct FiniMsg;
 
+type ObjectGetterArg = Box<dyn ObjectGetter + Send + 'static>;
```
papertigers, line 422: So one thing about this design that would be nice to change is to not clutter the production code with the test code.  I think @rhb2 mentioned this on our walk through.  I think a better approach may be to have the Trait defined and then define your type and implement the trait.  Then in the test code define the File based getter and implement the trait. Then you could substitute the file getter for your real moray getter and test the code that way. This is the approach I took in cfwlogd: https://github.com/joyent/firewall-logger-agent/blob/master/cfwlogd/src/events.rs#L276-L292

rjloura: The issue is that we need to create a unique "getter" for each thread.  The getter also stores state which is unique to each thread.  So we could have a Vec<getters> and look each one up in each thread.  That would make thread pooling more cumbersome.  What I ended up doing was, first of all, renaming getter to "resolver".  Then I made a new trait (`ObjectResolverBuilder`), the implementations of which will create a per thread `ObjectResolver` for us.  That builder/creator can then be added to the `FileGenerator` struct as a field and have it's method called whenever we need to create a new resolver (for each thread).  The `ObjectResolver` is still responsible for getting the full object metadata record on a per object ID basis.

This allows us to completely separate the test and production code.

manager/src/jobs/evacuate.rs
rhb2: What I'm asking is the inner `json_obj` variable... can that not be named something else for clarity?  Especially since as you pointed out, it's not the same thing as the one on line 501.

rjloura: Changed.

rjloura:
I removed file locking.  We can only lock files if they are writable, and I'd rather not open them in the rebalancer as writable. My thinking is that the operator should set the file permissions as read only.  When/If this moves to the next phase we can have the scan job adjust the permissions once it is done writing to the file.

PR #132, opened 2020-08-18
MANTA-5319 rebalancer manager should report version in HTTP response header

rjloura:


manager/src/main.rs
```diff
@@ -81,6 +82,23 @@ struct UpdateJobParams {
     uuid: String,
 }
 
+pub fn get_version() -> String {
```
rjloura, line 85: Unfortunately, even though almost an exact copy of this function is in the `agent/src/main.rs file` as well we can't re-use it, because it depends on the environment for the crate/workspace at compile time.  So if we move it to a common place (i.e. the rebalancer workspace) it will simply report the name and version number of that workspace.

rhb2 (dismissed):

rhb2 (dismissed):

PR #136, opened 2020-08-21
MANTA-5383 rebalancer should provide some backpressure to sharkspotter

rjloura:


manager/src/jobs/evacuate.rs
```diff
@@ -3518,39 +3635,56 @@ fn metadata_update_broker_dynamic(
             loop {
                 if let Ok(msg) = update_rx.try_recv() {
                     debug!("Received metadata update message: {:#?}", msg);
-                    update_dynamic_metadata_threads(
+                    if let Err(e) = update_dynamic_metadata_threads(
                         &mut pool,
-                        &queue,
+                        dwm_tx.clone(),
```
rhb2, line 3640: Could we say `Arc::clone(&dwm.tx)`?

manager/src/jobs/evacuate.rs
```diff
@@ -3518,39 +3635,56 @@ fn metadata_update_broker_dynamic(
...
+
+                        info!(
+                            "MD Update: no data from assignment checker \
+                             thread: {}",
                             e
                         );
+
                         break;
                     }
                 };
```
rhb2, line 3682: Just verifying my understanding, but the only difference is that we're using channel instead of the queue?

manager/src/jobs/evacuate.rs
```diff
@@ -924,6 +939,73 @@ impl EvacuateJob {
             "Assignment cache reduced by {}",
             initial_size - assignment_cache_usage(&assignments)
         );
+
+        let (lock, cv) = &*self.assignment_full_cv;
+        let _ = lock.lock().expect("cv lock");
+        cv.notify_one();
```
rhb2, line 945: I could be way off the reservation in my understanding but instead of unconditionally signaling, could we just only signal if we are below the high water mark?

rjloura:
Closing this PR for now.  It seems this is not an issue at the scale we are operating at.  It is something that should be done, but given the current constraints it is not worth the risk or time to get it integrated.

PR #138, opened 2020-08-26
MANTA-5389 rebalancer build should reflect correct postgres version

rjloura:
MANTA-5390 rebalancer zone should have psql in the root user PATH

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1224,7 +1224,7 @@ dependencies = [
 
 [[package]]
 name = "manager"
-version = "0.2.2"
+version = "0.3.0"
 dependencies = [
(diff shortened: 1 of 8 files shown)


timfoster (changes requested):
This looks good, but I think we also need changes in Makefile.postgres to set 
`PG12DIR = /opt/postgresql/12.4` there too

rjloura:
> This looks good, but I think we also need changes in Makefile.postgres to set
> `PG12DIR = /opt/postgresql/12.4` there too

ah good eye.  Fixed.

timfoster (approved):
Thanks for those changes, lgtm

PR #147, opened 2020-09-28
MANTA-5419 want rebalancer retry job

rjloura:


manager/src/jobs/evacuate.rs
```diff
@@ -3398,10 +3562,6 @@ fn metadata_update_one(
         );
     } else {
         let md_update_time = now.elapsed().as_micros();
-        job_action.md_update_time.fetch_add(
```
rjloura, line 3401: I removed this because we aren't using it anymore, and ounces make pounds.

rhb2 (approved):

